PR #17, opened 2024-02-14
feat: add api

butonium:


.buildkite/collect-bonds.yml
```diff
@@ -0,0 +1,50 @@
...
+    concurrency: 1
+
+  - wait: ~
+
+  - label: ":microscope: Collect Bonds"
+    commands:
+    - 'buildkite-agent artifact download --include-retried-jobs target/release/collect .'
+    - 'chmod +x target/release/collect'
+    - |
+      ./target/release/collect \
```
janlegner, line 32: this binary name is a bit too generic - especially as it there is other collect jobs, the same goes for  `store`

I suggest to convert collect to validator-bonds-cli or similar (this could be used to get other data from chain or do whatever on-chain).

I also suggest to make store a new binary inside the api crate.

Then if you change api's data::data to repositories::bond, all functions related to loading/storing DB data of the bonds would go there; the new binary in the api crate would call the code inside the crate::repositories::bond to store the data - making for a bit more concise code

store/src/commands/bonds.rs
```diff
@@ -0,0 +1,140 @@
...
+use tokio_postgres::types::ToSql;
+use tokio_postgres::NoTls;
+
+use crate::commands::utils::{InsertQueryCombiner, UpdateQueryCombiner};
+
+use super::common::CommonStoreOptions;
+
+const DEFAULT_CHUNK_SIZE: usize = 500;
+
+pub async fn store_bonds(options: CommonStoreOptions) -> anyhow::Result<()> {
```
janlegner, line 14: I know where you copied this from but I believe this can be done in a better way (now that i know rust a bit better). Inspiration could be taken from here:
https://github.com/marinade-finance/native-staking/blob/master/bot/src/api/handlers/rewards_metrics.rs#L70

I would approve the PR even without refactor of this section (first please make sure that the business logic is in place). But afterwards please make sure a follow-up PR is made and the `UpdateQueryCombiner` is no longer used.

Cargo.toml
```diff
@@ -3,7 +3,9 @@ members = [
     "programs/*",
     "insurance-engine",
     "snapshot-parser",
-    "merkle-tree"
+    "merkle-tree",
+    "cli",
```
janlegner, line 7: still a bit generic - wdyt @ochaloup ?

api/src/bin/api.rs
```diff
@@ -0,0 +1,80 @@
...
+        .and(with_context(context.clone()))
+        .and_then(bonds::handler);
+
+    let routes = top_level
+        .or(route_api_docs_oas)
+        .or(route_api_docs_html)
+        .or(route_bonds)
+        .with(cors);
+
+    warp::serve(routes).run(([0, 0, 0, 0], 8000)).await;
```
ochaloup, line 71: This is a nitpick and possibly a personal perspective. I appreciate when I have chance to configure the port without a need of changing code. The default `8000` is fine but when some other service during development is running it could be a clash. Some evn property for this maybe?

cli/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+anchor-spl = { workspace = true }
+chrono = { workspace = true }
+solana-clap-utils = { workspace = true }
+solana-client = { workspace = true }
+solana-logger = { workspace = true }
+solana-cli-config = { workspace = true }
+solana-account-decoder = { workspace = true }
+solana-program = { workspace = true }
+rust_decimal = { workspace = true }
+validator-bonds = { version = "0.1.0", path="../programs/validator-bonds", features=["cpi"] }
```
ochaloup, line 34: maybe this could be done in the same way as @janlegner configured the `merkle-tree` dependency?
https://github.com/marinade-finance/validator-bonds/blob/main/Cargo.toml#L32

cli/src/commands/bonds.rs
```diff
@@ -0,0 +1,57 @@
...
+use serde_yaml;
+use solana_account_decoder::UiAccountEncoding;
+use solana_client::{
+    rpc_config::{RpcAccountInfoConfig, RpcProgramAccountsConfig},
+    rpc_filter::Memcmp,
+};
+use validator_bonds::state::bond::Bond;
+
+const BONDS_DISCRIMINATOR: [u8; 8] = [224, 128, 48, 251, 182, 246, 111, 196];
+const BONDS_PROGRAM_ID: &str = "vBoNdEvzMrSai7is21XgVYik65mqtaKXuSdMBJ1xkW4";
```
ochaloup, line 13: probably not necessary while possible to load the program id from `validator-bonds` when it's as a dependency
https://github.com/marinade-finance/validator-bonds/blob/main/programs/validator-bonds/src/constants.rs#L4

the reason why it's duplicated in contract code is that there is (at least at anchor 0.29.0 what I believe) a way to get the program id to anchor IDL. So there is already duplication of the program ids in contract as it's in `lib.rs` as well
https://github.com/marinade-finance/validator-bonds/blob/main/programs/validator-bonds/src/lib.rs#L31
but I added at least a cargo test to get it failed when something changes

cli/src/commands/bonds.rs
```diff
@@ -0,0 +1,57 @@
...
+use anchor_lang::AccountDeserialize;
+use serde_yaml;
+use solana_account_decoder::UiAccountEncoding;
+use solana_client::{
+    rpc_config::{RpcAccountInfoConfig, RpcProgramAccountsConfig},
+    rpc_filter::Memcmp,
+};
+use validator_bonds::state::bond::Bond;
+
+const BONDS_DISCRIMINATOR: [u8; 8] = [224, 128, 48, 251, 182, 246, 111, 196];
```
ochaloup, line 12: The bonds discriminator could be taken from dependency code and not needed to be hardcoded.
I believe it should be possible to do like

```
use validator_bonds::state::bond::Bond;
use use anchor_lang::Discriminator;

// use in code
let d = Bond::DISCRIMINATOR;
```

Cargo.toml
ochaloup: Agree. It could be some more descriptive name. There is already an TS CLI for admin operations under `/packages`.

This name could be something with `bot` or `ci` or `db`, maybe?

ochaloup (commented):
@butonium do you think there could be some short `README` under `/cli` and `/api` how is expected the code to be run during development, testing and fixing things?

Cargo.toml
```diff
@@ -27,16 +28,30 @@ bincode = "1.3.3"
...
 merkle-tree = { path = "./merkle-tree" }
+validator-bonds = { path = "./programs/validator-bonds" }
 serde = "1.0.190"
 serde_json = "1.0.108"
 snapshot-parser = { path = "./snapshot-parser" }
 solana-client = "=1.16.18"
 solana-ledger = "=1.16.18"
 solana-program = "=1.16.18"
 solana-runtime = "=1.16.18"
+solana-account-decoder = "1.16.18"
```
janlegner, line 43: nitpick: this was sorted before

PR #18, opened 2024-02-15
Update to 1.17.22 + share lib code

ochaloup:
This update the mono repo solana dependency to 1.17.22 (current version on mainnet).
And moving the `TreeNode` to be shared under `merkle-tree` module. With that changing the `solana-sdk` to `solana-program` as it's dependency needed for contracts.

I would be happy to hear any feedback how to better structure the project if this module splitting solution does not fit.
Plus, I'm currently not able to oversee if the change of the solana version may do some issues elsewhere.

PR #19, opened 2024-02-16
[merkle tree] using Pubkey instead of string

ochaloup:
Refactoring on using Pubkey as base for TreeNode merkle tree generation instead of using string type.

Reasons:

* the TreeNode we expect to use consists of pubkeys. The verification of contract constraints work with pubkeys. There are no (or there could not be) reasons to convert types here and there.
* the string conversions consumes CU for contract. As a difference with and without string conversion for contract could be.

ad 1) This PR depends on merge of the PR: https://github.com/marinade-finance/validator-bonds/pull/18
ad 2) I don't know about consequences of changes here from string to pubkey in merkle tree generation. At least I assume the old generated merkle trees that are on GCP would be incompatible with the new one. Not sure if it can be a blocker of this change.

```
# string conversion
Program vBoNdEvzMrSai7is21XgVYik65mqtaKXuSdMBJ1xkW4 consumed 196170 of 1000000 compute units
# string conversion with CU tuning
Program vBoNdEvzMrSai7is21XgVYik65mqtaKXuSdMBJ1xkW4 consumed 121097 of 200000 compute units

# pubkey usage
Program vBoNdEvzMrSai7is21XgVYik65mqtaKXuSdMBJ1xkW4 consumed 71967 of 1000000 compute units
# pubkey usage with CU tuning
Program vBoNdEvzMrSai7is21XgVYik65mqtaKXuSdMBJ1xkW4 consumed 62309 of 200000 compute units
```


insurance-engine/src/insurance_claims.rs
```diff
@@ -27,10 +29,21 @@ pub struct InsuranceClaimCollection {
     pub claims: Vec<InsuranceClaim>,
 }
 
-pub fn stake_authorities_filter(whitelist: HashSet<String>) -> Box<dyn Fn(&StakeMeta) -> bool> {
+pub fn stake_authorities_filter(whitelist: HashSet<Pubkey>) -> Box<dyn Fn(&StakeMeta) -> bool> {
     Box::new(move |s| whitelist.contains(&s.stake_authority))
 }
 
+pub fn stake_authorities_filter_string(
```
janlegner, line 36: Why do we need this with _string? we can just pass Pubkeys in (parse Pubkeys in the argument of the cli directly)

janlegner (approved):

insurance-engine/src/insurance_claims.rs
ochaloup: Ok, I haven't realized. I was thinking that I need some manual conversion to Pubkey but it's about implementing the `FromStr` trait. PR updated, https://github.com/marinade-finance/validator-bonds/pull/19/commits/07b6236332d66ee7fada4cdfe9c0c37c47a62f81

PR #24, opened 2024-03-01
Mint bond configure

ochaloup:
This should be the last PR before contract audit. Still missing to finish comment cleaning. But the functionality should be okeyish for the audit review (believing...).

The new feature:
* configure bond with mint token

Fixes:
* multiple refactoring


ochaloup:
Comments refactored and hopefully clean. Missing piece is reworking the mint bond configure instruction that has flaws on invalidating the existing tokens. Under discussion with Neodyme on how to do.

PR #30, opened 2024-03-29
Updated settlement and merkle tree generation

janlegner:


scripts/generate-discord-public-report.bash
```diff
@@ -0,0 +1,82 @@
...
+if (( $(<"$settlement_collection_file" jq '.settlements | length' -r) == 0 ))
+then
+    echo "No settlements in epoch $epoch."
+    exit
+fi
+
+decimal_format="%0.9f"
+
+function fmt_human_number {
+    numfmt --to si $@
```
ochaloup, line 23: I didn't know this command. I need to add to my toolbox.

scripts/generate-discord-public-report.bash
```diff
@@ -0,0 +1,82 @@
...
+echo "                                vote account    settlement                   reason   stake     funded by"
+echo "--------------------------------------------+-------------+------------------------+-------+-------------"
+while read -r settlement
+do
+    vote_account=$(<<<"$settlement" jq '.vote_account' -r)
+    claims_amount=$(<<<"$settlement" jq '.claims_amount / 1e9' -r | xargs printf $decimal_format)
+    protected_stake=$(<<<"$settlement" jq '[.claims[].active_stake] | add / 1e9' -r | xargs -I{} bash -c 'fmt_human_number "$@"' _ {})
+    reason_code=$(<<<"$settlement" jq '.reason | keys[0]' -r)
+
+    if ! [[ $reason_code == "ProtectedEvent" ]]
```
ochaloup, line 38: I'm not clear if that's worthy but somehow I feel it would be nice to have the "constant" somehow documented. I mean in the Rust code to have some note at the struct definition that's the name is used in pipeline bash script?
Does it sound reasonable or not?

.buildkite/prepare-claims.yml
```diff
@@ -47,37 +47,40 @@ steps:
         --output-validator-meta-collection "$$snapshot_dir/validators.json" \
         --output-stake-meta-collection "$$snapshot_dir/stakes.json"
 
+  - label: ":floppy_disk: :arrow_left: :cloud: Fetch past validator data"
+    env:
+      gs_bucket: gs://marinade-validator-bonds-mainnet
+    commands:
+    - 'snapshot_dir=$(buildkite-agent meta-data get snapshot_dir)'
+    - 'epoch=$(buildkite-agent meta-data get epoch)'
+    - 'previous_epoch=$((epoch - 1))'
```
ochaloup, line 56: When the previous epoch did not generate the `validators.json` - is it something we expect it can happen or it's just something that "won't happen"? If possibly happens it seems a manual intervention is needed, could that be mentioned somewhere what could be the process?

ochaloup (approved):
All seems nice to me. I added few comments while not sure how relevant they are.

settlement-config.yaml
```diff
@@ -0,0 +1,24 @@
+---
+- LowCreditsSettlement:
+    meta:
+      funder: ValidatorBond
+    min_settlement_lamports: 100000000
+    grace_low_credits_bps: 100 # 1 %
+    covered_range_bps: [0, 2000] # 0 % - 20 %
```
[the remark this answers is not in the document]
janlegner: Let me explain on example:
Expected rewards 100 SOL, downtime 20 %, actual rewards will be 80 SOL.
The range [0, 20 %] covers at most 0.2 * 100 SOL, in this case 20 SOL.
The range [20%, 100%] would start covering if downtime was above 20 %. If it was 20 % exactly, it would have to cover exactly 0 SOL.

I am not sure it makes sense to talk about inclusion/exclusion in this case; If it does, then I think, strictly speaking, both bounds are included - but on the boundary, one of the settlements will be 0

scripts/generate-discord-public-report.bash
janlegner: It does make sense, at the same time pretty much all fields of all collections are already used by other processes after those collections are stored in the buckets. The names should not change - we will have to document all places which would be broken by this...

.buildkite/prepare-claims.yml
janlegner: This is expected when we are starting the process in the past to backfill data or for the first ever run of this updated pipeline. Previous data not existing is therefore error only sometimes. For cases other than backfill, this will not happen as it would mean the previous run did not succeed and it was not fixed (which is unexpected).

All that is to be done is re-run the failing pipeline, not sure this is worth documenting.

settlement-config.yaml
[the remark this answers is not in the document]
ochaloup: I see, thanks for the elaboration

.buildkite/prepare-claims.yml
ochaloup: Ok, got it, thanks I did not have this context.

PR #34, opened 2024-04-17
[pipelines] cli bot command for settlement creation

ochaloup:
The idea is to get the json with merkle trees and creating the settlements accounts for particular epoch.
It could be the command is run several times with the same input json and thus CLI verifies existence of the settlement accounts on-chain and base of that it put transaction execution in-place or not.

### DIFF
--- a/.buildkite/init-settlements.yml
+++ b/.buildkite/init-settlements.yml
@@ -5,11 +5,12 @@ steps:
   # epoch number provided in env or prompted
   - input: "What is the epoch to init settlements for?"
     fields:
-      - text: "Epoch number"
+      - text: "Epoch"
         key: "epoch"
+        format: "[0-9]+"
     if: "build.env('EPOCH') == null"
 
-  - command: "echo 'Snapshot dir: $$EPOCH'"
+  - command: "echo 'Epoch is: $$EPOCH'"
     if: "build.env('EPOCH') != null"
   - block: "Confirm to get processed init settlements"
     prompt: "Please review the 'Unblock' if ready to proceed."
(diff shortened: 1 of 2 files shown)


ochaloup:
After today talk I'm switching this to draft. I miss to straighten the builder library and fix tests. When ready I will switch it back to PR and will ask for review on discord channel / @janlegner

ochaloup:
The init settlement cli and buildkite pipeline is ready for review.

settlement-pipelines/src/bin/init_settlement.rs
```diff
@@ -0,0 +1,747 @@
...
+            .to_str().ok_or(anyhow!("Cannot convert file name {} to string", args.input_merkle_tree_collection))?;
+        let re = Regex::new("^([0-9]+)").unwrap();
+        let captures = re.captures(merkle_file_name);
+        captures
+            .and_then(|c| c.get(0))
+            .and_then(|m| m.as_str().parse::<u64>().ok())
+            .ok_or(
+                anyhow!("--epoch not provided and cannot extract epoch number from input merkle tree collection file path '{}'", args.input_merkle_tree_collection)
+            )
+    }, Ok)?;
```
janlegner, line 205: Not needed as the Collection has epoch inside (but let this be a note for the future iteration, not to change now)

settlement-pipelines/src/bin/init_settlement.rs
```diff
@@ -0,0 +1,747 @@
...
+                        Some(stake_accounts_to_fund.remove(0))
+                    };
+                if let Some(FundBondStakeAccount {
+                    stake_account: destination_stake,
+                    split_stake_account: destination_split_stake,
+                    ..
+                }) = stake_account_to_fund
+                {
+                    info!(
+                        "Settlement:{}, merging stake accounts to {}",
```
janlegner, line 512: Merging seems unnecessary but let's keep it here

settlement-pipelines/src/bin/init_settlement.rs
```diff
@@ -0,0 +1,747 @@
...
+                            amount_to_fund,
+                            lamports_available
+                        );
+                    } else if lamports_available > amount_to_fund + minimal_stake_lamports {
+                        // we are in situation that the stake account has got (or it will have after merging)
+                        // more lamports than needed for funding the settlement in current loop iteration,
+                        // the rest of lamports from the stake account can be used for other settlements,
+                        // the lamports will be available under split stake account after the stake account is funded (next for-loop section)
+                        // let's use the split stake account as a source for funding of next settlement
+                        // WARN: this REQUIRES that the fund bond transactions are executed in sequence!
```
janlegner, line 555: But if we did not merge now, it wouldn't...

janlegner (approved):

settlement-pipelines/src/bin/init_settlement.rs
ochaloup: I realized that. Thanks for pointing it out

settlement-pipelines/src/bin/init_settlement.rs
ochaloup: Hm... maybe I was a bit overthinking. Maybe the merge process would be better to be in part of the cleaning pipeline where rent is gathered or similar. Or merging belongs rather to part of claiming the most probably.

My thinking was to have prepared stake account that's prepared for any claiming and the claiming pipeline does not need to be interested in. But  that's not right. It should be the claim that merges when necessary.
This pipeline should be without merge and thus simpler. Considered for refactor.

PR #38, opened 2024-04-24
[common-rs] update methods

ochaloup:
Update of common-rs with bonds methods loading data from chain.

### DIFF
--- a/validator-bonds-cli/src/commands/bonds.rs
+++ b/validator-bonds-cli/src/commands/bonds.rs
@@ -4,6 +4,7 @@ use serde_yaml;
 use solana_sdk::pubkey::Pubkey;
 use std::str::FromStr;
 use std::sync::Arc;
+use validator_bonds_common::constants::MARINADE_CONFIG_ADDRESS;
 use validator_bonds_common::funded_bonds::collect_validator_bonds_with_funds;
 
 pub async fn collect_bonds(options: CommonCollectOptions) -> anyhow::Result<()> {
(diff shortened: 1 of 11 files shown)


ochaloup:
@janlegner : I pushed few more changes based on the `clippy` reporting. But still ready for review.

On top of that I would like know if we can change the `solana_sdk::stake::state::StakeState` for `StakeStateV2` or if there is some issue to stay with the `StakeState`?

```
warning: use of deprecated enum `solana_sdk::stake::state::StakeState`: Please use `StakeStateV2` instead, and match the third `StakeFlags` field when matching `StakeStateV2::Stake` to resolve any breakage. For example, `if let StakeState::Stake(meta, stake)` becomes `if let StakeStateV2::Stake(meta, stake, _stake_flags)`.
  --> common-rs/src/stake_accounts.rs:10:26
   |
10 |     stake::{self, state::StakeState},
   |                          ^^^^^^^^^^
   |
   = note: `#[warn(deprecated)]` on by default
```

ochaloup:
Update of the lib on finding locked stake accounts.

validator-bonds-cli/src/commands/bonds.rs
```diff
@@ -12,7 +13,7 @@ pub async fn collect_bonds(options: CommonCollectOptions) -> anyhow::Result<()>
         options.commitment.to_string(),
     ));
 
-    let config_address = Pubkey::from_str("vbMaRfmTCg92HWGzmd53APkMNpPnGVGZTUHwUJQkXAU")?;
+    let config_address = Pubkey::from_str(MARINADE_CONFIG_ADDRESS)?;
```
janlegner, line 16: uf, forgot about this one :/

janlegner (approved):

PR #44, opened 2024-04-29
Pipelines settlement reporting

ochaloup:
Changing the settlements pipeline to print reports of running the commands to discord notification.
Currently nothing much around formatting.


.buildkite/init-settlements.yml
```diff
@@ -2,49 +2,47 @@ agents:
...
             "url": "'"$$BUILDKITE_BUILD_URL"'",
             "color": "8388863"
           }
         ]
       }'
 
-  - label: ":gear: Setup"
+  - block: "Confirm to get processed init settlements"
+    prompt: "Please review the 'Unblock' if ready to proceed."
+    if: "build.env('EPOCH') != null"
```
janlegner, line 36:
```suggestion
```
I think this is just a copy-paste error, right?

janlegner (approved):

.buildkite/init-settlements.yml
ochaloup: No, it was meant to get blocked processing when init settlements `EPOCH` env variable is provided, i.e., in case of the other pipeline triggers this one.

When the `EPOCH` env is not provided, i.e., this is executed by hand from buildkite, then there is prompt instead of that where the user has to provide the epoch number.

janlegner: ah - otherwise it would be obtained from `buildkite-agent meta-data get epoch`, i see

PR #45, opened 2024-04-30
Pipeline settlement adjust

ochaloup:
Enriching info messages in pipeline processing. Finding amount of funded from number of stake accounts from chain for any type of funding.


janlegner (approved):

ochaloup:
from the review added one more test check, removing pipeline yml testing data and will merge. thanks.

PR #46, opened 2024-04-30
Pipelines settlement claiming

ochaloup:
Settlement claiming pipeline. 

~~(Unfortunately, ended up with a spaghetti again. I will need to refactor when all settles.)~~
~~This is untested (at least from base) and I'm creating it as draft with in hope will be finished during tomorrow.~~


janlegner:
> I will need to refactor when all settles

Pun intended? :)

janlegner (approved):

PR #48, opened 2024-05-02
[settlement-pipelines] fixes on claiming

ochaloup:
* fix way to find out the deactivated stake accounts
* using loaded stake accounts despite they are not all
* TS test alignments


ochaloup:
Just FYI @janlegner but going to merge :shrug:

PR #52, opened 2024-05-06
[settlement-pipelines] claim prioritization

ochaloup:
Changeset for:
* checking if target accounts are locked
* looking for stake accounts by stake/withdraw authority one by one instead of looking through all the stake accounts existing on-chain. I tried running that (https://buildkite.com/marinade-finance/claim-settlements/builds/25#018f4eda-e126-4d45-8c13-c4a7e2a7deaa) and this concept brings few timeouts on RPC but only a minor number. Otherwise seems working good.

The processing contains listing of `SettlementClaim` accounts with not `getProgramAccounts` but with `getMultipleAccounts`. It developed that before that way so placing here as a topic for discussion. Maybe just using getProgramAccounts is better. Just I was thinking the asking for account addresses is easier for RPC than querying the program accounts by filter. That's why.

Let me know what you think.


janlegner:
`getProgramAccounts` is not better as those accounts are from 2 epochs back and can be merged/not existing/split; we specifically want to find current stake accounts;

there are 3 stake authorities for which we need to find all stake accounts, so let's use the `getProgramAccounts` -> if we can do that in Native Staking, we can sure make it work here....

ochaloup:
> `getProgramAccounts` is not better as those accounts are from 2 epochs back and can be merged/not existing/split; we specifically want to find current stake accounts;
> 
> there are 3 stake authorities for which we need to find all stake accounts, so let's use the `getProgramAccounts` -> if we can do that in Native Staking, we can sure make it work here....

There is maybe misunderstanding. The code uses the `getProgramAccounts` for stake accounts as we dicussed - it searches all stake accounts of the stake/withdrawer pair when needed and those searches are cached.
The getMultipleAccounts is used for `SettlementClaim` accounts where the addresses can be precalculated from the data available in json merkle tree files.

ochaloup:
Updating PR to contain only the prioritization code and not mixing it with unfinished reporting improvements (another PR will be created).

settlement-pipelines/src/stake_accounts.rs
```diff
@@ -0,0 +1,69 @@
...
+pub fn prioritize_for_claiming(
+    stake_accounts: &CollectedStakeAccounts,
+    clock: &Clock,
+    stake_history: &StakeHistory,
+) -> anyhow::Result<Pubkey> {
+    let mut non_locked_stake_accounts = stake_accounts
+        .iter()
+        .filter(|(_pubkey, _, stake)| !is_locked(stake, clock))
+        .collect::<Vec<_>>();
+    non_locked_stake_accounts.sort_by(|(_, _, stake_account1), (_, _, stake_account2)| {
```
janlegner, line 23: sort_by_cached_key would make way more sense here

settlement-pipelines/src/bin/claim_settlement.rs
```diff
@@ -358,59 +339,54 @@ async fn main() -> anyhow::Result<()> {
                 }
             };
 
-            if stake_accounts_per_staker_cache
-                .get(&tree_node.stake_authority)
-                .is_none()
+            let stake_withdraw_pair = StakeWithdrawAuthorityPair::new(
```
janlegner, line 342: it would be much niceer if we had a `struct StakeAccountsCache` and you would just call `.get(witdraw_authority, stake_authority)`
then you would call `pick_stake_account_for_claim` which would give you account or tell you that only locked accounts are in there

settlement-pipelines/src/stake_withdraw_pair.rs
```diff
@@ -0,0 +1,31 @@
...
+impl<'a> StakeWithdrawAuthorityPair<'a> {
+    pub fn new(stake_authority: &'a Pubkey, withdraw_authority: &'a Pubkey) -> Self {
+        Self {
+            stake_authority,
+            withdraw_authority,
+        }
+    }
+}
+
+impl PartialEq for StakeWithdrawAuthorityPair<'_> {
```
janlegner, line 17: why not just
```rust
#[derive(PartialEq, Eq, Hash)]
pub struct StakeWithdrawAuthorityPair<'a> {
    pub stake_authority: &'a Pubkey,
    pub withdraw_authority: &'a Pubkey,
}
```
?

ochaloup: Much better, thanks!

settlement-pipelines/src/stake_accounts.rs
ochaloup: :+1:  Not being aware of this std lib function. Thanks.

common-rs/src/stake_accounts.rs
```diff
@@ -213,7 +214,7 @@ fn map_stake_accounts_to_settlement(
             }
         }
     }
-    // sorting stake accounts that the delegated ones are before the non-delegated ones
+    // sorting stake accounts for the delegated ones being before the non-delegated ones
```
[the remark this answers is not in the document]
ochaloup: This is a bit of left-over that I decided to leave here when the functionality will be enhanced.

The idea behind is that when the settlement could be funded from two sources - from validator bond and from marinade then the claiming should prioritize the validator bond source. Marinade funds Initialized non delegated stake accounts so that should be moved at the end of the list.

Anyway, not needed not, it should be rather removed as makes rather a code buzz here.

settlement-pipelines/src/bin/claim_settlement.rs
ochaloup: In full agreement, thanks a lot for pointing that out.

ochaloup:
@janlegner I updated the PR based on your comments.

I think I have only a bit trouble to implement nicely the suggestion about the `StakeAccountsCache`. To my consideration to make it nicer in code I need some abstraction around reporting. It's something I started to work on but I haven't got finished that yet.
If you meant to do that in different way to be more straightforward than I did, I will be glad to know.

janlegner:
It is fine, we can improve iteratively, no need to fix everything in a single PR

PR #55, opened 2024-05-15
[settlement-pipeline] rent payer for init and claim

ochaloup:
Init and claim settlements to use different wallets for tx fee payer and rent payer.
For close settlements I assume there will be used the `VALIDATOR_BONDS_MARINADE_TX_FEE_PAYER` to pay for tx fees.

### DIFF
--- a/.buildkite/claim-settlements.yml
+++ b/.buildkite/claim-settlements.yml
@@ -67,12 +67,15 @@ steps:
     - '. "$HOME/.cargo/env"'
     - 'buildkite-agent artifact download --include-retried-jobs target/release/claim-settlement .'
     - 'chmod +x target/release/claim-settlement'
+    - 'echo "UNKNOWN ERROR" > ./claiming-report.txt'
     - |
       ./target/release/claim-settlement \
         --rpc-url $$RPC_URL \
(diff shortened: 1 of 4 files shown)


.buildkite/claim-settlements.yml
```diff
@@ -67,12 +67,15 @@ steps:
     - '. "$HOME/.cargo/env"'
     - 'buildkite-agent artifact download --include-retried-jobs target/release/claim-settlement .'
     - 'chmod +x target/release/claim-settlement'
+    - 'echo "UNKNOWN ERROR" > ./claiming-report.txt'
```
janlegner, line 70: leftover?

janlegner (approved):

.buildkite/claim-settlements.yml
ochaloup: Yes, or maybe rather an error. Idea behind was to have reported an `unknown error` when something wrong happens on call claim settlement before it goes running and to ensure there is the artifact existing for further use. But as the `tee` is creating new file either way this should be ensured. Leaving this as it is as it is changed in follow-up PR.
Thanks for the feedback here.

PR #59, opened 2024-05-31
[pipelines] init to verify funding stake state

ochaloup:
The `init_settlement` pipeline started to fail as the pipeline wrongly did not consider the stake account state and was not able to merge in such cases.

The change verifies the stake state and lets only those with the same state merge. If the state differs then it funds the accounts directly without merging beforehand.

Still, the merging is beneficial for claiming we need the best one big account that is just split into parts and not pointing into settlements multiple accounts that need to stay with "overflow" of 1 SOL as being the requirements for the existence of the stake account.


ochaloup:
I decided to refactor a bit to be a little easier :shrug:  to read (taken out functionality to a function, not wrapping vec and option together when we can use empty vec instead of `None`).

The functionality is the same so I will merge it.

PR #66, opened 2024-06-20
feat: add max stake wanted

butonium:


api/src/repositories/bond.rs
```diff
@@ -84,6 +85,7 @@ pub async fn store_bonds(options: CommonStoreOptions) -> anyhow::Result<()> {
             params.push(Box::new(epoch));
             params.push(Box::new(bond.updated_at));
             params.push(Box::new(bond.cpmpe));
+            params.push(Box::new(bond.max_stake_wanted));
```
janlegner, line 88: PARAMS_PER_INSERT must be increased

ochaloup (approved):
All seems ok to me.

PR #74, opened 2024-07-09
[contract] fund makes possible when activating

ochaloup:
Changing `fund_bond` instruction to permit to funding of any stake account that is:
* delegated 
* non-locked

Before this change, it does matter if it's activating or deactivating etc. It was permitted to get funded only fully activated stake account.

The idea behind the change is to make it easier to fund a bond for validators.
It should be possible to create a stake account and immediately fund it to the bond account without the need to wait a whole epoch to get it activated (as it is for a newly created stake account right now).

The technical presumption of the change is 

* The `fund_bond` changes the `withdraw` and `staker` authorities to a PDA `validator bonds authority` based on the program id and config address.
* The bond-funded stake account can be used to fund the Settlement. The `fund_settlement` instruction checks locking, delegation and authorities and then deactivates the stake.
* For just bond-funded stake to be used for settlement on a protected event or bidding then:

  1. The stake account is used in the next (or follow-up) epochs from the time the stake account was funded - for settlement 
funding, it's the same situation as bond funding only fully activated stake accounts.
  2. The stake account is used in the current epoch, in the same epoch as it was funded to bond - deactivation (by a matter of calling `fund_settlement`) happens when the stake is in an activating state which makes it deactivated and usable for claiming.

_Note: the current implementation does not limit the possibility of having the stake account being activated/deactivated only partially. There are timeouts of 3 epochs when Settlement is active and can be funded and can be claimed. It is "expected" that the deactivation is finished by Solana until the Settlement is closed. The deactivated stake accounts can be withdrawn._

### DIFF
--- a/packages/validator-bonds-sdk/__tests__/bankrun/bankrun.ts
+++ b/packages/validator-bonds-sdk/__tests__/bankrun/bankrun.ts
@@ -6,15 +6,22 @@ import {
 import { Keypair, PublicKey } from '@solana/web3.js'
 import {
   BankrunExtendedProvider,
+  currentEpoch,
   testInit,
   warpToNextEpoch,
 } from '@marinade.finance/bankrun-utils'
-import { delegatedStakeAccount } from '../utils/staking'
+import {
+  StakeStates,
+  delegatedStakeAccount,
+  getAndCheckStakeAccount,
+} from '../utils/staking'
 import {
   executeFundBondInstruction,
   executeInitBondInstruction,
 } from '../utils/testTransactions'
 import 'reflect-metadata'
+import { BN } from 'bn.js'
+import { U64_MAX } from '@marinade.finance/web3js-common'
 
 export async function initBankrunTest(programId?: PublicKey): Promise<{
   program: ValidatorBondsProgram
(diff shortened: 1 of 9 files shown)


packages/validator-bonds-sdk/__tests__/bankrun/bankrun.ts
```diff
@@ -98,3 +119,54 @@ export async function delegateAndFund({
...
+      'activationEpoch',
+      activationEpoch.toString(),
+      'deactivationEpoch',
+      deactivationEpoch.toString(),
+      'currentEpoch',
+      curEpoch.toString()
+    )
+
+    if (!deactivationEpoch.eq(U64_MAX) && deactivationEpoch.gte(curEpoch)) {
+      return StakeActivationState.Deactivating
```
janlegner, line 156: activationEpoch = 500
deactivationEpoch = 500
---> for this case the correct state would be Deactivated

packages/validator-bonds-sdk/__tests__/bankrun/bankrun.ts
```diff
@@ -98,3 +119,54 @@ export async function delegateAndFund({
...
+    } else if (!activationEpoch.eq(U64_MAX) && activationEpoch.gte(curEpoch)) {
+      return StakeActivationState.Activating
+    } else if (
+      !deactivationEpoch.eq(U64_MAX) &&
```
janlegner, line 160: first half of the condition seems a bit superfluous

packages/validator-bonds-sdk/__tests__/bankrun/bankrun.ts
```diff
@@ -98,3 +119,54 @@ export async function delegateAndFund({
...
+      deactivationEpoch.lt(curEpoch)
+    ) {
+      return StakeActivationState.Deactivated
+    } else if (!activationEpoch.eq(U64_MAX) && activationEpoch.lt(curEpoch)) {
```
janlegner, line 164: dtto

programs/validator-bonds/src/checks.rs
```diff
@@ -180,6 +182,45 @@ pub fn check_stake_is_not_locked(
...
+) -> Result<Stake> {
+    if let Some(stake) = stake_account.stake() {
+        let StakeHistoryEntry {
+            effective,
+            activating,
+            deactivating,
+        } = stake
+            .delegation
+            .stake_activating_and_deactivating(epoch, Some(stake_history), None);
+        if deactivating > 0 || effective == 0 {
```
janlegner, line 202: this condition is also true for activating stake accounts

janlegner (approved):

PR #83, opened 2024-08-06
feat: add bid distribution

butonium:


bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,112 @@
...
+            for ((withdraw_authority, stake_authority), stake_metas) in grouped_stake_metas {
+                if !stake_authority_filter(stake_authority) {
+                    continue;
+                }
+                let stake_accounts: HashMap<_, _> = stake_metas
+                    .iter()
+                    .map(|s| (s.pubkey, s.active_delegation_lamports))
+                    .collect();
+                let active_stake: u64 = stake_accounts.values().sum();
+                let claim_amount = ((active_stake as f64 / effective_sam_stake as f64) * stakers_total_claim as f64) as u64;
```
ochaloup, line 75: This seems to me wrong but correct me when otherwise.

The `active_stake` will be equal but rather bigger than the `effective_sam_stake`. From that the `claim_amount` will be bigger than `stakers_total_claim`, won't be?

bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,112 @@
...
+                        stake_authority: **stake_authority,
+                        stake_accounts,
+                        claim_amount,
+                        active_stake,
+                    });
+                    claims_amount += claim_amount;
+                }
+            }
+
+            if claims_amount > 0 {
```
ochaloup, line 99: Maybe add some sanity checks to verify at the end of each cycle that the sum of claims_amount is not bigger than stakers_total_claim?

ochaloup (changes requested):
I have several questions and comments in the code.

Then please run the `pnpm lint:fix` (or fmt with clippy similar to `cargo fmt --all && cargo clippy --fix --allow-staged --allow-dirty`) to fix the code style.

.buildkite/prepare-bidding.yml
```diff
@@ -0,0 +1,83 @@
...
+        --marinade-fee-stake-authority stWirqFCf2Uts1JBL1Jsd3r6VBWhgnpdPxCTe1MFjrq \
+        --marinade-fee-withdraw-authority stWirqFCf2Uts1JBL1Jsd3r6VBWhgnpdPxCTe1MFjrq \
+        --output-settlement-collection "./${claim_type}-settlements.json" \
+        --output-merkle-tree-collection "./${claim_type}-settlement-merkle-trees.json"
+    artifact_paths:
+    - "./${claim_type}*.json"
+
+  - wait: ~
+  
+  - label: ":floppy_disk: :arrow_right: :cloud: Upload artifacts"
```
ochaloup, line 70: What about discord reports to psr-feed and our info channel?
https://github.com/marinade-finance/validator-bonds/blob/main/.buildkite/prepare-protected-events.yml#L96-L121

Or are there so many events that it's not what should be reported to feed? Maybe to report with sending a file to not spam the ds channel with many messages?

.buildkite/prepare-bidding.yml
```diff
@@ -0,0 +1,83 @@
...
+
+  - wait: ~
+
+  - label: ":floppy_disk: :arrow_left: :cloud: Downloading Settlements JSON data"
+    commands:
+    - |
+      epoch=${EPOCH:-$(buildkite-agent meta-data get epoch)}
+      buildkite-agent meta-data set epoch "$$epoch"
+    - |
+      curl -s "https://scoring.marinade.finance/api/v1/scores/sam?epoch=$$epoch" -o validators.json
```
ochaloup, line 40: I believe it would be nice to name the output name differently to validators.json. It's connected to SAM and it's provided to argument --sam-meta-collection so something with SAM. Like sam-scores or at least sam-validators (or rather sam-scores-validators).

.buildkite/prepare-bidding.yml
ochaloup, line 70: What about discord reports to psr-feed and our info channel?
https://github.com/marinade-finance/validator-bonds/blob/main/.buildkite/prepare-protected-events.yml#L96-L121

Or are there so many events that it's not what should be reported to feed? Maybe to report with sending a file to not spam the ds channel with many messages?

bid-distribution/src/sam_meta.rs
```diff
@@ -0,0 +1,61 @@
...
+use serde::Serializer;
+use serde::Deserializer;
+use {
+    merkle_tree::serde_serialize::pubkey_string_conversion,
+    serde::{Deserialize, Serialize},
+    std::fmt::Debug,
+};
+
+#[derive(Serialize, Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
```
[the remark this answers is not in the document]
butonium: I tried to use without it, but I get an error for field not found. (I guess there is no rename by default, it takes it as it is)

ochaloup (changes requested):
Two more comments. The discrepancy appeared as I started with the review yesterday evening and the code changed meanwhile while the in-progress review was pointing to a non-existent code.

bid-distribution/src/settlement_config.rs
```diff
@@ -0,0 +1,42 @@
+use serde::{Deserialize, Serialize};
+use settlement_engine::settlement_claims::SettlementMeta;
+use solana_sdk::pubkey::Pubkey;
+
+#[derive(Clone, Deserialize, Serialize, Debug)]
+pub enum SettlementConfig {
```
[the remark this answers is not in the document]
butonium: makes sense, will change it

bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,112 @@
...
+use std::cmp::min;
+use settlement_engine::settlement_claims::{SettlementReason, Settlement, SettlementClaim, SettlementCollection};
+
+pub fn generate_bid_settlement_collection(
+    stake_meta_index: &StakeMetaIndex,
+    sam_validator_metas: &Vec<ValidatorSamMeta>,
+    stake_authority_filter: &dyn Fn(&Pubkey) -> bool,
+    settlement_config: &SettlementConfig,
+) -> SettlementCollection {
+    let settlements = generate_bid_settlements(&stake_meta_index, &sam_validator_metas, &stake_authority_filter, &settlement_config);
```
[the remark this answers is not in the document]
butonium: I was thinking about it, but it will not be as straight forward as there. I guess I can extract epoch from sam_validator_metas vec and compare it with stake_meta epoch.

bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,112 @@
...
+            let marinade_stake: u64 = stake_meta_index.iter_grouped_stake_metas(&validator.vote_account).unwrap()
+                .filter(|(&(_, &stake_authority), _)| stake_authority_filter(&stake_authority))
+                .flat_map(|(_, metas)| metas.iter())
+                .map(|meta| meta.active_delegation_lamports)
+                .sum();
+
+            if sam_target_stake + mnde_target_stake == 0 {
+                continue;
+            }
+            let stake_sam_percentage = sam_target_stake as f64 / (sam_target_stake as f64 + mnde_target_stake as f64);
```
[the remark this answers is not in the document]
butonium: as far as I know there can be a case where marinade_stake is > sam_target_stake + mnde_target_stake. When a validator is outbid or the bond is reduced and is limiting the target stakes for next epochs but the validator was not unstaked

bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,112 @@
...
+            let initial_sam_stake =  (marinade_stake as f64 * stake_sam_percentage) as u64;
+            let effective_sam_stake: u64 = min(initial_sam_stake, max_wanted_stake);
+            let effective_total_bid_claim = (effective_sam_stake as f64 * validator.effective_bid) as u64;
```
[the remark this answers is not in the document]
butonium: > The `effective_bid` is somehow(?) calculated from the bid placed by the validator (i.e., what's available in bond account data `costPerMillePerEpoch`. That bid is of bidding how many lamports the validator wants to pay for 1000 of SOL. Is the fact that the bid is per 1000 SOLs somehow considered in the API providing the value and it was transferred to bid per 1 SOL?
> 
you're right, it should be converted to / 1 SOL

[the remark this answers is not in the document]
butonium: > The `effective_sam_stake` is the stake of related stake accounts (filtered by authorities, currently authorities belonging to Marinade) that is proportion to the stake being under the management of SAM. Such value is in SOL? But then there is the `max_wanted_stake` that goes from the `min` call where `max_wanted_stake` is in lamports, right? Isn't that a discrepancy?
> 

as far as I know `effective_sam_stake` is a percentage of `marinade_stake` which is computed as sum of `active_delegation_lamports` filtered by marinade authority.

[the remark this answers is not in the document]
butonium: so both are lamports values

bid-distribution/src/settlement_claims.rs
butonium: you're right, the ratio is reversed

bid-distribution/src/settlement_claims.rs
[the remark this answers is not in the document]
ochaloup: > so both are lamports values

Correct, you're right, I've got confused myself. Sorry.

bid-distribution/src/sam_meta.rs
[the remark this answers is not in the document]
ochaloup: I've just understood that I was wrong the default is `snake_case` and there is no (it seems) way to switch to default to be `camelCase`. "Strangely" the TS works by default with `snakeCase` 🤦

bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,168 @@
...
+        .map(|s| (s.pubkey, s.active_delegation_lamports))
+        .collect();
+
+    let mut settlement_claim_collections = vec![];
+
+    for validator in sam_validator_metas {
+        if let Some(grouped_stake_metas) =
+            stake_meta_index.iter_grouped_stake_metas(&validator.vote_account)
+        {
+            if validator.effective_bid == Decimal::ZERO {
```
ochaloup, line 68: I like this. I should fix the protected events processing in the same way (what I haven't done despite I check with you here, blame on me :-))

.buildkite/prepare-bidding.yml
```diff
@@ -0,0 +1,121 @@
...
+    commands:
+    - 'buildkite-agent artifact download --include-retried-jobs sam-scores.json .'
+    - 'buildkite-agent artifact download --include-retried-jobs stakes.json .'
+    - 'buildkite-agent artifact download --include-retried-jobs target/release/bid-distribution-cli .'
+    - 'chmod +x target/release/bid-distribution-cli'
+    - |
+      ./target/release/bid-distribution-cli \
+        --sam-meta-collection ./sam-scores.json \
+        --stake-meta-collection ./stakes.json \
+        --marinade-fee-bps 800 \
```
ochaloup, line 60: Another thing for my interest.
Here it's defined 8% of the fee for Marinade. When we calculate APY from Bonds bidding is the marinade fee subtracted from the APY that is presented at the dashboard? Could the fee be changed and thus the sync of the value here and what's presented could be good to be somehow double-checked or synchronized?

bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,168 @@
...
+                continue;
+            }
+            let sam_target_stake =
+                validator.marinade_sam_target_sol * Decimal::from_f64(1e9).unwrap();
+            let mnde_target_stake =
+                validator.marinade_mnde_target_sol * Decimal::from_f64(1e9).unwrap();
+            let max_wanted_stake = validator.max_stake_wanted * Decimal::from_f64(1e9).unwrap();
+            let marinade_payment_percentage =
+                Decimal::from(*settlement_config.marinade_fee_bps()) / Decimal::from(10000);
+            let effective_bid = validator.effective_bid / Decimal::from_f64(1e3).unwrap();
```
ochaloup, line 78: To me using the numeric format `1e3` is not readable as well as when there is used `1000`. But it may be only me.

ochaloup, line 78: Let me please double-check here. The `effective bid` is the bond configuration pmpe (`costPerMillePerEpoch`) stands for `how many lamports the validator wants to pay for 1000 SOLs`? Is that right?
Then `effective_bid` is "re-defined" by the calculation here as `number of lamports per 1 SOL`. Don't we calculate all other values in `lamports per lamport`?

bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,168 @@
...
+                withdraw_authority: *settlement_config.marinade_withdraw_authority(),
+                stake_authority: *settlement_config.marinade_stake_authority(),
+                stake_accounts: marinade_fee_deposit_stake_accounts.clone(),
+                claim_amount: marinade_fee_claim,
+                active_stake: marinade_fee_deposit_stake_accounts.values().sum(),
+            });
+            claims_amount += marinade_fee_claim;
+
+            assert!(
+                claims_amount <= stakers_total_claim + marinade_fee_claim,
```
ochaloup, line 153: Should not be here rather considered the `effective_total_claim` that should be more or less (that's and issue that because of calculation it is not equal probably?) equal to sum of `stakers_total_claim + marinade_fee_claim`?

bid-distribution/src/settlement_config.rs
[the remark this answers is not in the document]
ochaloup: This was probably not changed but I believe it is pretty minor.

bid-distribution/src/settlement_claims.rs
```diff
@@ -0,0 +1,168 @@
...
+}
+
+pub fn generate_bid_settlements(
+    stake_meta_index: &StakeMetaIndex,
+    sam_validator_metas: &Vec<ValidatorSamMeta>,
+    stake_authority_filter: &dyn Fn(&Pubkey) -> bool,
+    settlement_config: &SettlementConfig,
+) -> Vec<Settlement> {
+    info!("Generating bid settlements...");
+    let sam_validators_epoch = sam_validator_metas.first().unwrap().epoch;
```
ochaloup, line 41: Maybe to use `all()` instead of `first()`?

ochaloup (changes requested):
Thanks for addressing my points. Most of the comments are just minors except two,
to me it would be nice to upload a discord report to gcloud and I wonder around the calculation of effective bid if it's ok or not.

.buildkite/prepare-bidding.yml
butonium: that was a placeholder, I need to double check with lj or someone who knows what pct marinade will take. I think I can make it as a dynamic argument to the pipeline

bid-distribution/src/settlement_claims.rs
butonium: sure will make it 1000

butonium: so the thing is that effective_bid is taken from ds-sam pipeline result via scoring api (the format is in SOL, not lamports)

.buildkite/prepare-bidding.yml
ochaloup: Yeap, ok. Would you know about the other part how the APY is calculated? Is there considered the Marinade fee to reduce the APY for stakers in the dashboard?
Or is it probably rather a question on @janlegner , right?

butonium: I changed the `--marinade-fee-bps` after a short sync with him, but regarding APY not sure, I think he would know better.

bid-distribution/src/settlement_claims.rs
butonium: I wanted to double check that the sum of all claims doesn't exceed the effective_total_claim, which is indeed (stakers_total_claim + marinade_fee_claim)

bid-distribution/src/settlement_claims.rs
ochaloup: I got the point and I'm sorry for the confusion. Thanks for the elaboration you're right :pray:

bid-distribution/src/settlement_claims.rs
```diff
@@ -38,12 +38,13 @@ pub fn generate_bid_settlements(
...
 ) -> Vec<Settlement> {
     info!("Generating bid settlements...");
-    let sam_validators_epoch = sam_validator_metas.first().unwrap().epoch;
-
-    assert_eq!(
-        stake_meta_index.stake_meta_collection.epoch, sam_validators_epoch as u64,
-        "SAM Validators Collection epoch must be same as stake meta collection epoch"
-    );
+    
+    for sam_validator_meta in sam_validator_metas {
```
ochaloup, line 42: This is just a super nitpick just thinking how I read the code. When I can see the `for` loop it seems that something should happen, like some mutation.
When there is just an assert with some function it feels for the first grasp being just some assertion, like this:

```
    assert!(
        sam_validator_metas.iter().all(|v| v.epoch as u64 == stake_meta_index.stake_meta_collection.epoch), 
        "SAM Validators Collection epoch must be same as stake meta collection epoch"
    );
    ```

**Review by ochaloup — APPROVED** [review]
Thanks. To me it is all good.

File: bid-distribution/src/settlement_claims.rs (hunk 15 above)
Comment (butonium) [reply bid-distribution/src/settlement_claims.rs:42]: you're right, that way is better

File: .buildkite/prepare-bidding.yml (hunk 11 above)
Comment (janlegner) [reply .buildkite/prepare-bidding.yml]: > Yeap, ok. Would you know about the other part how the APY is calculated? Is there considered the Marinade fee to reduce the APY for stakers in the dashboard?

This is TBD, the same way MNDE directing or re-balancing costs are not shown in the dasboard

**Comment by janlegner** [comment]
@butonium please update the pipeline in bk to reflect that this branch has been merged

File: bid-distribution/src/settlement_claims.rs [hunk 16]
```diff
@@ -0,0 +1,169 @@
...
+            let sam_target_stake =
+                validator.marinade_sam_target_sol * Decimal::from_f64(1e9).unwrap();
+            let mnde_target_stake =
+                validator.marinade_mnde_target_sol * Decimal::from_f64(1e9).unwrap();
+            let max_wanted_stake = validator.max_stake_wanted * Decimal::from_f64(1e9).unwrap();
+            let marinade_payment_percentage =
+                Decimal::from(*settlement_config.marinade_fee_bps()) / Decimal::from(10000);
+            let effective_bid = validator.effective_bid / Decimal::from(1000);
+
+            if sam_target_stake + mnde_target_stake == Decimal::ZERO {
```
janlegner, line 81: this is not correct @butonium @ochaloup 

case:
- validator A already had 1 M SOL stake from Marinade
- for whatever reason, the next auction run shows the target for the validator is 0
- we still charge bid in this case!

janlegner: https://github.com/marinade-finance/ds-sam/blob/main/src/commands/analyze-revenue.cmd.ts#L145

PR #86, opened 2024-08-13
[protected event dist] fixing issue on bps commission f64 calculation

ochaloup:


ochaloup:
We discussed with @butonium . He reviewed and we agreed it should be all changed to get used the decimal than f64.

PR #87, opened 2024-08-14
fix: avoid adding empty claims

butonium:


bid-distribution/src/settlement_claims.rs
```diff
@@ -142,28 +128,45 @@ pub fn generate_bid_settlements(
...
+                    x.withdraw_authority
+                        .eq(settlement_config.marinade_withdraw_authority())
+                        && x.stake_authority
+                            .eq(settlement_config.marinade_stake_authority())
+                })
+                .iter()
+                .map(|s| (s.pubkey, s.active_delegation_lamports))
+                .collect();
+
+            if effective_sam_stake > 0 && marinade_fee_claim > 0 {
```
ochaloup, line 145: Is the `effective_sam_stake` correct here? It's calculated from max stake wanted and not from the stake that is available in the stake accounts (i.e., `active_stake`).
(I would like to understand the meaning of the check)

ochaloup (commented):
I'm not sure about the `active_stake` check and the value provided to `SettlementClaim` `active_stake` value. Otherwise all fine from me.

bid-distribution/src/settlement_claims.rs
butonium: > Is the `effective_sam_stake` correct here? It's calculated from max stake wanted and not from the stake that is available in the stake accounts (i.e., `active_stake`). (I would like to understand the meaning of the check)

my bad it should be `initial_sam_stake` instead. It's not the `total_active_stake` because we should use only the SAM part of the total stake accounts delegated to this validator, since MNDE part is not paying bids.

butonium: > Is the `effective_sam_stake` correct here? It's calculated from max stake wanted and not from the stake that is available in the stake accounts (i.e., `active_stake`). (I would like to understand the meaning of the check)

sync on discord, please check last changes

ochaloup (approved):
+1 from me

PR #88, opened 2024-08-14
Non sam events compatibity fix

ochaloup:
I found an issue around backwards compatibility and the ability to work with old JSON data within the project.
The closing, funding etc. needs to be able to work with old and new data (I started to use v1 - not with sam and v2 - with some, but if there is better naming please propose)

With that the protected event of new name is `CommissionSamIncrease` and placing back the other types of protected events from v1.


butonium (approved):
lgtm

PR #92, opened 2024-08-21
fix: duplicate active stake

butonium:


janlegner:
The hardcoded fee address is not great -> this is quite hidden... I am considering other options before giving approval

janlegner (approved):

PR #95, opened 2024-08-29
Fix for funding bin cli pipeline

ochaloup:
The PR covers small refactoring of naming, adding a little bit of logging info for the fund pipeline to have better insight into what happened and mostly fixes an issue of failing pipeline on funding when the same stake account could be taken for two different settlement of the same vote account.  As for wrong calculation in the pipeline, it could happen that after funding there is not enough left for the financing of the other. When both such fundings are placed into the same transaction the transaction fails and nothing is funded.

The issue fix is the change of the if statement
https://github.com/marinade-finance/validator-bonds/compare/pipelines-more-info-logging?expand=1#diff-ed4510b215e1a2a295acd1b99614e0b3b95c79d5fcea050225b08cd2afcad716L400

### DIFF
--- a/protected-event-distribution/src/settlement_claims.rs
+++ b/protected-event-distribution/src/settlement_claims.rs
@@ -1,4 +1,5 @@
 #![allow(clippy::type_complexity)]
+
 use crate::{
     protected_events::ProtectedEvent,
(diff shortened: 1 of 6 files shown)


butonium (approved):
lgtm

PR #96, opened 2024-08-30
Adjusting buildkite pipelines

ochaloup:
Two things adjusted here:

1. Prepare bidding/protected-events pipeline to fund only newly created settlements (avoiding race conditions on not existing settlements from in parallel running pipeline of the other type)
2. PSR feed is updated only after initialization is executed (not immediately after JSON data is created and not before an operator manually permits init settlement processing)


butonium (approved):
lgtm

PR #98, opened 2024-09-03
[pipelines] fix wrong claim type for funding pipelinel

ochaloup:
I made an error in processing buildkite metadata processing. It works with init settlement but I forgot to check the claiming.

https://buildkite.com/docs/agent/v3/cli-meta-data#setting-data-description
```
buildkite-agent meta-data set <key> [value] [options...]

Set arbitrary data on a build using a basic key/value store.

You can supply the value as an argument to the command, or pipe in a file or script output.
```

and **!**

```
The value must be a non-empty string, and strings containing only whitespace characters are not allowed.
```

Follow-up to #96

### DIFF
--- a/.buildkite/claim-settlements.yml
+++ b/.buildkite/claim-settlements.yml
@@ -24,7 +24,7 @@ steps:
   - label: ":black_nib: Find Discord webhook"
     commands:
     - |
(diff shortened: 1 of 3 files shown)


butonium (approved):
lgtm

PR #113, opened 2024-09-13
[snapshot parse] fixing way to get mev commission info

ochaloup:
A fix for the wrong way of parsing MEV commission from the snapshot.

The reasoning to start with the fix is the following:

. The bonds parser takes snapshot data from Jito gcloud from the last slot of the epoch.
. The `TipDistribution` account is created by a validator when it gets a leader slot and creates a block and could get some tips.
. At the end of the epoch, all the validators processed their slots and all data has to be on the chain
. What is not on-chain are the Merkle tree root data that is placed into the particular accounts by Jito at the start of the following epoch
. Bonds parser searches for MEV commission data already part of the blockchain. There is no reason to work with `epoch - 1`.


butonium (approved):
lgtm

PR #115, opened 2024-09-24
[pipeline] seq executor run one by one with error returning

ochaloup:
Depending on passing PR https://github.com/marinade-finance/solana-transaction-builder/pull/6


ochaloup:
Using approved dependency from `solana-transaction-builder` as of `main#bd448f8f7350313b1c61548e01895ecf6050930c"`

image

PR #140, opened 2024-11-12
[pipeline] working with warnings when no stake account exists

ochaloup:
Add warning error handling that won't fail the pipeline but it reports them to a report data.


butonium (approved):

butonium:
lgtm

PR #157, opened 2025-01-07
[pipelines] using stake accounts with biggest non-delegated part first on settlement funding

ochaloup:
The perspective of this change is to re-define the way stake accounts are considered for funding-created settlements.
Up to now (with the context of having the issue in the contract for some time until October 2024 - https://github.com/marinade-finance/validator-bonds/pull/77) the funding pipeline was taking the biggest stake account to be used for funding the settlement.

This change prioritised using the stake accounts with the biggest part of undelegated lamports. Those can be directly used for claiming the settlements while the split part is left delegated and not claimed.

### DIFF
--- a/settlement-pipelines/src/stake_accounts.rs
+++ b/settlement-pipelines/src/stake_accounts.rs
@@ -65,14 +65,14 @@ pub enum StakeAccountStateType {
 }
 
 pub fn get_stake_state_type(
-    stake_account: &StakeStateV2,
+    stake_account_state: &StakeStateV2,
     clock: &Clock,
     stake_history: &StakeHistory,
 ) -> StakeAccountStateType {
-    if let StakeStateV2::Initialized(_) = stake_account {
+    if let StakeStateV2::Initialized(_) = stake_account_state {
         // stake account is initialized and not delegated, it can be delegated just now
         StakeAccountStateType::Initialized
-    } else if let Some(delegation) = stake_account.delegation() {
+    } else if let Some(delegation) = stake_account_state.delegation() {
         // stake account was delegated, verification of the delegation state
         let StakeHistoryEntry {
             effective,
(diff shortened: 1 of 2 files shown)


settlement-pipelines/src/stake_accounts.rs
```diff
@@ -97,6 +97,23 @@ pub fn get_stake_state_type(
...
+    clock: &Clock,
+    stake_history: &StakeHistory,
+) -> u64 {
+    if let Some(delegation) = stake_account_state.delegation() {
+        let StakeHistoryEntry {
+            effective,
+            deactivating,
+            activating,
+        } = delegation.stake_activating_and_deactivating(clock.epoch, Some(stake_history), None);
+        effective + deactivating + activating
```
martinkrecek, line 111: (non-blocking)
I'd need to dive a lot deeper into the context to understand/decide on my side whether such optimization would make sense, so I'll just share my thought for you to either look into or ignore (both is perfectly fine with me) and we can have a separate chat/call if needed.

TLDR: Are you sure that both `deactivating` and `activating` should be included in the sum here, i.e. neither of these two stake parts can be used for the funding?

The `effective` part of stake is clear to me, we don't want to touch that, but I'm thinking that the other two `(de)activating` parts might potentially be useful. Depending on the time spans in the settlements flows (there are some delays in between the steps), we could consider treating `(de)activating` stake as non-delegated (inactive) too in this context. Because it might happen that the stake part will either be deactivated just in time or it won't be activated yet anyway, I'm trying to understand whether it couldn't be optimized so that some of these stake parts would also be utilized.

martinkrecek (approved):
Looks good to me, approving with one idea/topic for discussion in a comment.

settlement-pipelines/src/stake_accounts.rs
ochaloup: This is a good point. I would separate different aspects here.

a) Thanks for the review. I urge it as it is to check the next epoch of the behaviour.
b) The goal here is to take all parts of delegated lamports. The code then calculates the un-delegated amount from that. There are used all three parts as those are lamports that cannot be directly withdrawn - i.e., they cannot be immediately claimed. When Settlement is funded with immediately available funds the claiming does not need to wait for the end of the epoch when the `deactivating` happens to be `inactive`.
With funding, it happens that the whole amount (whole funded stake account) is moved to deactivated by the funding instruction.
c) The relevant part to your point is that the `inactive` lamports are moved for claiming while the delegated part is left under Bond management (to be used for later funding possibly). Here some optimization could be considered as `deactivating` part could be considered as `inactive` as it will be either way deactivated.
On the other hand the `activating` should be considered as `active` in my opinion.
d) Maybe the method could be named better to respect the fact the `deactivating` part is part of the returned amount.

TLDR; as you mentioned it's not blocking I will check the behaviour in the pipeline and will get back to you for further discussion when I will got some insight on how that works (and if I find time to inspect in more details)

janlegner:
> These are my stakes these funds are mine and it's not allowing me to withdraw or claim or cli unstake and withdraw it's always some thing. I don't understand I apologize where I erroring. Support is not so supportive.

Did you create support ticket? If so, which number?

PR #158, opened 2025-01-07
[cli] claim-withdraw-request: adding option to claim for particular stake account

ochaloup:
A user reported an issue with not being able to claim his withdrawal request.

The trouble is caused by a small issue in the contract that does not permit to split stake account that has got a bigger number of undelegated lamports. Funding of settlements suffered with this issue and the contract fix for claiming will be similar to the approach in PR  https://github.com/marinade-finance/validator-bonds/pull/77

This PR adds a workaround for the user who reported the issue to force claiming with a specific stake account. This workaround should help this particular user but it's not a generic fix.


martinkrecek (approved):
Looks good to me, just two small typos and a double-check comment.

ochaloup:
I updated the version to 2.0.5, published the CLI and going to merge this. @martinkrecek thanks.

PR #159, opened 2025-01-08
[contract] fix claim withdraw request splitting

ochaloup:
As for the issue of non-delegated lamports troubles the funding settlement of the same issue was left as part of the claim withdrawal request.

This is a minor issue as it does not permit the withdraw request in some cases (if the stake account has got a bigger number of undelegated lamports). A workaround is to delete the withdraw request and create a new one with a bigger amount.

This PR adds the logic of withdrawing non-delegated lamports into the split account before calling split instruction.

Missing parts:

* tests for claim withdraw request with non-delegated lamports
* considering if the changed ix signature should not be made of version V2 and leaving the old signature within the directory `v1` for the ix would be parsable in future

### DIFF
--- a/packages/validator-bonds-sdk/generated/validator_bonds.ts
+++ b/packages/validator-bonds-sdk/generated/validator_bonds.ts
@@ -1061,6 +1061,16 @@ export type ValidatorBonds = {
           "isSigner": false
         },
         {
+          "name": "rent",
+          "isMut": false,
+          "isSigner": false
+        },
+        {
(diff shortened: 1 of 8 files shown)


ochaloup:
The workaround to this functionality flaw is to cancel existing withdraw request and create new with an amount that will follow the size of the undelegated lamports.

This fix (when elaborated and PR finished) is expected to be considered in some later time when a new deployment of the contract is planned.

PR #161, opened 2025-01-13
Fix claim withdraw request

ochaloup:


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## TS CLI&SDK [2.0.6](https://github.com/marinade-finance/validator-bonds/compare/v2.0.5...v2.0.6) (2025-01-13)
```
martinkrecek, line 5: The comparison link doesn't lead to a valid diff (the same for `2.0.5` below), based on a quick check I'd say that it's due to a missing tag `v2.0.5` (and `v2.0.6` likely too).

martinkrecek (approved):

PR #165, opened 2025-01-21
add trivy scanner

shejby:


ochaloup (commented):
The github workflow files seem all ok. But the job running the scan fails
```
Running Trivy with options: trivy fs .
2025-01-21T14:01:21Z	ERROR	[vulndb] The first run cannot skip downloading DB
2025-01-21T14:01:21Z	FATAL	Fatal error	init error: DB error: database error: --skip-update cannot be specified on the first run
```

Is that expected?

ochaloup (approved):

PR #175, opened 2025-02-04
Version bumps and fixing script for publish

ochaloup:
I did a wrong publish of npm package. I believe it was caused by the fact of not re-building the old version of the package on my local system. That's why I changed the package.json script and needed a double update of versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,19 +2,21 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## TS CLI&SDK [2.1.2](https://github.com/marinade-finance/validator-bonds/compare/v2.1.0...v2.1.2) (2025-02-04)
-
-### Fixes
-
-Fix of wrong npm registry deployment
-
-## TS CLI&SDK [2.1.0](https://github.com/marinade-finance/validator-bonds/compare/v2.0.6...v2.1.0) (2025-02-04)
+## TS CLI&SDK [2.1.2](https://github.com/marinade-finance/validator-bonds/compare/v2.1.1...v2.1.2) (2025-02-04)
 
 ### Updates
 
 * cli: `fund-bond-sol` subcommand provides a way to fund bond directly from the user's wallet without a need to manually create a stake account
 * cli institutional: creating a new package `@marinade.finance/validator-bonds-cli-institutional` where functionality is based on
   the original `validator-bonds-cli` while streamlining the experience to focus strictly on managing bond accounts
+
+## TS CLI&SDK [2.1.1](https://github.com/marinade-finance/validator-bonds/compare/v2.1.0...v2.1.1) (2025-02-04)
+
+Wrong version, skipped.
+
+## TS CLI&SDK [2.1.0](https://github.com/marinade-finance/validator-bonds/compare/v2.0.6...v2.1.0) (2025-02-04)
+
+Issue with published deployment.
 
 ## TS CLI&SDK [2.0.6](https://github.com/marinade-finance/validator-bonds/compare/v2.0.5...v2.0.6) (2025-01-13)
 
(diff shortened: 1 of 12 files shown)


packages/validator-bonds-cli-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@marinade.finance/validator-bonds-cli-core",
-  "version": "2.1.0",
+  "version": "2.1.2",
```
martinkrecek, line 3: (non-blocking)
Just curious why it's bumped to `2.1.2` and not `2.1.1`? 🤔

pnpm-lock.yaml
```diff
@@ -68,10 +68,10 @@ importers:
         specifier: '=2.4.13'
         version: 2.4.13
       '@marinade.finance/validator-bonds-cli-core':
-        specifier: 'workspace: ^2.1.0'
+        specifier: 'workspace: ^2.1.1'
```
martinkrecek, line 71: All the packages are bumped to `2.1.2`, shouldn't that be reflected here in the lock too?

CHANGELOG.md
```diff
@@ -2,7 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## TS CLI&SDK [2.1.0](https://github.com/marinade-finance/validator-bonds/compare/v2.0.6...v2.1.0) (2025-01-31)
+## TS CLI&SDK [2.1.2](https://github.com/marinade-finance/validator-bonds/compare/v2.1.0...v2.1.2) (2025-02-04)
```
martinkrecek, line 5: Flagging that neither of the tags `v2.1.0` and `v2.1.2` exists, and adding a note that I'm not sure if I should be checking this in future PRs, because afaik it's usually tagged only after merging/publishing.

CHANGELOG.md
```diff
@@ -2,7 +2,13 @@
...
+
+### Fixes
+
+Fix of wrong npm registry deployment
+
+## TS CLI&SDK [2.1.0](https://github.com/marinade-finance/validator-bonds/compare/v2.0.6...v2.1.0) (2025-02-04)
```
martinkrecek, line 11: Is the date changed from `2025-01-31` to `2025-02-04` for this older log on purpose?

packages/validator-bonds-cli-core/package.json
ochaloup: The trouble with 2.1.1 is of wrong published version in CLI. Thus it needed to be re-published :shrug:

pnpm-lock.yaml
ochaloup: Yes, thanks. updated

CHANGELOG.md
ochaloup: I updated the changelog a bit. Let's see if it is better now.

CHANGELOG.md
ochaloup: Yes, that was an errror of the changelog data. The deployment was considered to happen but as the PR was not reviewed last week I did not publish it. So it is just aligning the doc with reality.

CHANGELOG.md
```diff
@@ -2,14 +2,22 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## TS CLI&SDK [2.1.0](https://github.com/marinade-finance/validator-bonds/compare/v2.0.6...v2.1.0) (2025-01-31)
+## TS CLI&SDK [2.1.2](https://github.com/marinade-finance/validator-bonds/compare/v2.1.1...v2.1.2) (2025-02-04)
```
martinkrecek, line 5: (non-blocking)
There is now no description of this new version (it previously said "Fix of wrong npm registry deployment" in this PR), not sure if intended.

PR #178, opened 2025-02-18
[MIP.10] removal of stake wanted, increase TVL cap to 4%

ochaloup:
Implementing MIP.10 proposal

* https://forum.marinade.finance/t/mip-10-simplification-of-sams-delegation-strategy/1713
* https://app.realms.today/dao/MNDE/proposal/E5HsJzBZhniUfoRrT7aV1gAJ2VXqFQqvwWp8AcgSco9i

Removal of max stake wanted from the CLI (configure, show, readme)

Notes:
* The existence of the commented-out text for the CLI arguments is intentional. The contract still supports this configuration option and maybe later it can be added back. While it seems wrong from a review perspective I'm not such strong in this opinion.
* the dependencies update works for the local development. For publishing there needs to be added a new commit without `workspace:`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
-## TS CLI&SDK [2.1.3](https://github.com/marinade-finance/validator-bonds/compare/v2.1.2...v2.1.3) (2025-02-20)
+## TS CLI&SDK [2.1.3](https://github.com/marinade-finance/validator-bonds/compare/v2.1.2...v2.1.3) (2025-02-19)
 
 ### Updates
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## TS CLI&SDK [2.1.3](https://github.com/marinade-finance/validator-bonds/compare/v2.1.2...v2.1.3) (2025-02-20)
+
+### Updates
+
+* cli: `configure-bond`, `init-bond`, `show-bond` removal of `--max-stake-wanted` parameter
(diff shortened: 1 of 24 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+## TS CLI&SDK [2.1.3](https://github.com/marinade-finance/validator-bonds/compare/v2.1.2...v2.1.3) (2025-02-20)
```
martinkrecek, line 5: (non-blocking)
The last tag in this repo that GH shows me is `v2.0.6`, if these diffs are expected to be used by the readers the tags should be added.

packages/validator-bonds-cli/src/commands/show.ts
```diff
@@ -37,7 +43,25 @@ export function installShowBond(program: Command) {
...
         })
       },
     )
 }
+
+export function reformatBondBidding(
+  key: string,
+  value: unknown,
+): ReformatAction {
+  if (!getCliContext().logger.isLevelEnabled('debug')) {
```
martinkrecek, line 56: (non-blocking)
I think I understand the logic of this, it removes `maxStakeWanted` in non-debug (non-local) envs, but it would maybe deserve a short comment as changing the behavior based on logger setup is not the cleanest (although I understand that it's a lot simpler than introducing some specific flags and does the job, so I'm not suggesting to change the logic). Ideally, the comment should explain why in local/debug usage the property `maxStakeWanted` is kept but in other envs it's removed (probably something like ~"removing it completely is not the best approach but we don't want to scare/confuse the users by printing 'nonexistent' properties"), because in effect this logic actually "hides" something from the output of the CLI (but the data contains it), so we should be careful about future maintenance/understanding.

martinkrecek (dismissed):

CHANGELOG.md
ochaloup: Thank you. Good catch. There is not an automation for the release notes and I need to do that manually.
It would be nice to have it in place for github pipeline but I haven't got time for it.

PR #181, opened 2025-02-28
Claim withdraw request fix + repo refactoring

ochaloup:
Consisting two parts

* refactoring of the repository structure to not troublesome issues with pnpm publish 
* fix for claim withdraw request to permit merging only fully active stake accounts

### DIFF
--- a/scripts/update-cli-version.sh
+++ b/scripts/update-cli-version.sh
@@ -3,37 +3,39 @@
 SCRIPT_PATH=`readlink -f "$0"`
 SCRIPT_DIR=`dirname "$SCRIPT_PATH"`
 
-SDK_PACKAGE_JSON="$SCRIPT_DIR/../packages/validator-bonds-sdk/package.json"
-CLI_PACKAGE_JSON="$SCRIPT_DIR/../packages/validator-bonds-cli/package.json"
-CLI_CORE_PACKAGE_JSON="$SCRIPT_DIR/../packages/validator-bonds-cli-core/package.json"
-CLI_INSTITUTIONAL_PACKAGE_JSON="$SCRIPT_DIR/../packages/validator-bonds-cli-institutional/package.json"
-PIPELINES_PACKAGE_JSON="$SCRIPT_DIR/../settlement-pipelines/package.json"
 CLI_INDEX="$SCRIPT_DIR/../packages/validator-bonds-cli/src/index.ts"
+CLI_INSTITUTIONAL_INDEX="$SCRIPT_DIR/../packages/validator-bonds-cli-institutional/src/index.ts"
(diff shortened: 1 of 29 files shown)


scripts/update-cli-version.sh
```diff
@@ -3,37 +3,39 @@
...
+SDK_PACKAGE_JSON="$SCRIPT_DIR/../packages/validator-bonds-sdk/package.json"
+[ ! -f "$SDK_PACKAGE_JSON" ] && echo "$SDK_PACKAGE_JSON not found" && exit 1
+PREVIOUS_VERSION=`cat $SDK_PACKAGE_JSON | grep version | cut -d '"' -f 4`
+PREVIOUS_ESCAPED_VERSION=$(echo $PREVIOUS_VERSION | sed 's/\./\\./g')
+
+# update package.json minor version
+for I in "$SCRIPT_DIR/../packages/"*sdk* "$SCRIPT_DIR/../packages/"*cli*; do
+  echo "Package: $I"
+  cd "$I"
+  pnpm version patch
```
mwhyte-mrnd, line 21: Do we automate only patch version updates? I mean, do we follow semver or any update is considered as patch?

tsconfig.json
```diff
@@ -1,14 +1,20 @@
...
+    "module": "commonjs",
     "esModuleInterop": true,
     "resolveJsonModule": true,
+    "incremental": true,
+    "strict": true,
     "skipLibCheck": true,
     "sourceMap": true,
     "experimentalDecorators": true,
     "emitDecoratorMetadata": true,
+    "noImplicitAny": true,
```
mwhyte-mrnd, line 14: When turning `strict: true` it also turns on `noImplicitAny` under the hood, as far as I remember.

mwhyte-mrnd (approved):
Although I am not familiar with this project in-depth, changes of this PR seem to be fine. Few notes:
- I see you use `ts-node` and just want to warn you that some of their issues/bugs are kind of ignored by their dev team for a long time and if you will encounter some of them in the future, you can try `tsx` as alternative.
- good job on `publish:*` pnpm scripts, nice showcase of how `--filter` and `-r` can be used to make scripts/config smaller

tsconfig.json
ochaloup: Ok. Thanks for noting it, I didn't know and haven't studied. I think we can leave the configuration with `noImplicitAny` as well being set to true as being more explicit on this configuration option.
(if you are not strictly against with a knowledge of a good practice)

scripts/update-cli-version.sh
ochaloup: Yes we are and I'm aware of this missing piece - let's say. When there is a minor update (or major even) that needs to be updated manually (as of now).

PR #182, opened 2025-03-01
[pipeline] close settlement pipeline checks for delegated stake

ochaloup:
There was an error when closing the settlement in the buildkite pipeline script. The contract expects to work with stake that is delegated but the pipeline was missing to filter that type of account.

[1] https://github.com/marinade-finance/validator-bonds/blob/main/programs/validator-bonds/src/instructions/settlement/close_settlement.rs

### DIFF
--- a/settlement-pipelines/src/settlements.rs
+++ b/settlement-pipelines/src/settlements.rs
@@ -225,8 +225,9 @@ pub async fn obtain_settlement_closing_refunds(
             } else {
                 split_rent_refund_accounts?
             };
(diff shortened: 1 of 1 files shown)


mwhyte-mrnd (approved):
As far as I understood we don't want first element, but we want the one which has delegation. Looks fine to me.

ochaloup:
> As far as I understood we don't want first element, but we want the one which has delegation. Looks fine to me.

Yes, correct. Non delegated account fails constraints of the contract.
Thanks for the review.

PR #186, opened 2025-03-13
Institutional calculation transformed into Settlements and used in on-chain pipelines [GEN-1344] [GEN-1310]

ochaloup:
* new CLI to create Settlement and Merkle Tree data from the institutional calculation
* changing Discord notifications to Slack notifications
* implementing buildkite processing for institutional calculation
* ~~Adjusting settlement on-chain creation to work with institutional settlements (and Bond config of institutions)~~ (will be delivered in another PR)


ochaloup:
@martinkrecek there are two main areas of the PR. One is adding a CLI program that generates Merkle tree from data of institutional staking calculations. Second is refactoring of buildkite to a) use Slack, b) be prepared for institutional staking type. 

I will be testing the changes on Monday where I need to verify if my buildkite refactoring has not caused any harm. It is difficult to test the buildkite processing without really running it on buildkite with real data.

(note: this PR will be followed - when the refactoring functionality is tested - by another PR that will change on-chain processing to allow working with institutional native staking proxy pubkeys)

ochaloup:
Sorry for changing under your hands. But there won't be more (I promise).

I realized I should add one small change that updates the link from the slack to point directly to the executed step. It is easier to navigate as the report is printed at the step.

.buildkite/prepare-institutional-distribution.yml
```diff
@@ -0,0 +1,140 @@
...
+    artifact_paths:
+      - "./institutional-payouts.json"
+
+  - wait: ~
+
+  - label: ":scales: Prepare Validator Bonds Merkle Tree"
+    env:
+    commands:
+    - 'buildkite-agent artifact download --include-retried-jobs institutional-payouts.json .'
+    - 'chmod +x target/release/institutional-distribution-cli'
```
martinkrecek, line 60:
```suggestion
    - buildkite-agent artifact download --include-retried-jobs institutional-payouts.json .
    - chmod +x target/release/institutional-distribution-cli
```
I see the quotes removed in other pipelines, so suggesting here too for consistency.

.buildkite/prepare-institutional-distribution.yml
```diff
@@ -0,0 +1,140 @@
...
+    - |
+      ./target/release/institutional-distribution-cli -- \
+        --institutional-payouts institutional-payouts.json \
+        --marinade-fee-stake-authority ${MARINADE_FEE_STAKE_AUTHORITY} \
+        --marinade-fee-withdraw-authority ${MARINADE_FEE_WITHDRAW_AUTHORITY} \
```
martinkrecek, line 65: As opposed to `prepare-bid-distribution.yml`, the "build level" env vars `MARINADE_FEE_(STAKE/WITHDRAW)_AUTHORITY` are not defined in this pipeline, I guess they were supposed to be copied here?

.buildkite/init-settlements.yml
```diff
@@ -215,16 +226,20 @@ steps:
...
+  - label: ":gear: Upload and get running fund-settlements pipeline"
     commands:
       - epoch=$(buildkite-agent meta-data get epoch)
       - claim_type=$(buildkite-agent meta-data get claim_type)
-      - 'echo "Epoch: $$epoch, Claim type: $$claim_type"'
+      - slack_feed=$(buildkite-agent meta-data get slack_feed)
+      - export EPOCH="$$epoch"
+      - export CLAIM_TYPE="$$claim_type"
+      - export SLACK_FEED="$$slack_feed"
+      - 'echo "Epoch: $$epoch, Claim type: $$claim_type, Slack Feed: $$slack_feed"'
```
martinkrecek, line 242:
```suggestion
      - echo "Epoch: '$$epoch', Claim Type: '$$claim_type', Slack Feed: '$$slack_feed'"
```
(non-blocking - minor inconsistency)

martinkrecek (commented):
Mostly small notes, possibly one case of missing env vars, otherwise looks ok. 👍

.buildkite/init-settlements.yml
```diff
@@ -25,41 +28,60 @@ steps:
...
       epoch=${EPOCH:-$(buildkite-agent meta-data get epoch)}
       buildkite-agent meta-data set epoch "$$epoch"
     - |
       claim_type=${CLAIM_TYPE:-$(buildkite-agent meta-data get claim_type)}
       buildkite-agent meta-data set claim_type "$$claim_type"
-    - 'echo "Epoch: $$epoch", Claim Type: $$claim_type'
+    - |
+      slack_feed=${SLACK_FEED:-$(buildkite-agent meta-data get slack_feed)}
+      buildkite-agent meta-data set slack_feed "$$slack_feed"
+    - 'echo "Epoch: $$epoch", Claim Type: $$claim_type, Slack Feed: $$slack_feed"'
```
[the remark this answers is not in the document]
ochaloup: There is an issue in buildkite  - that I'm not sure why it is so - but the simple `echo` does not work and fails. At least for my experience and processing. I will be happy to hear back what can be done better here.

But this fails:

```
- echo "Something $VAR"
``` 

And these passes

```
- 'echo "Something $VAR"'
```

```
-  "echo Something $VAR"
```

```
- |
   echo "Something $VAR"
```

.buildkite/init-settlements.yml
ochaloup: A note on this above.

ochaloup:
@martinkrecek PR updated and commented.

.buildkite/init-settlements.yml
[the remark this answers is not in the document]
martinkrecek: Whaaaat??? 😮 Ok, no comment, no idea how something like this can happen. 🤷‍♂️ And resolving, since apparently it's a workaround to a Buildkite bug and thus needed. ✅

PR #196, opened 2025-03-21
[settlement-pipelines][GEN-1260] claim for institutional with different priority

ochaloup:
Changing the claiming order of stake accounts for institutional staking.

I was considering to proceed with some CLI argument but then ended with using the same strategy as for liquid staking. Let's see if it is considered feasible.


ochaloup:
@janlegner, based on our discussion, I updated the claim priority for institutional staking
Now it is `DelegatedAndDeactivated -> Initialized -> DelegatedAndDeactivating -> DelegatedAndActive -> DelegatedAndActivating ...`

janlegner (approved):

PR #204, opened 2025-04-02
[codama] update SDK versions to work with @solana/kit

ochaloup:
Update the SDK for the Anchor program of Validator Bonds for @solana/kit. There is used the codama (https://github.com/codama-idl/codama) to generate the TS binding of the Anchor program. Solana library was named `@solana/web3.js` for version `2.0.0`. Now it is slightly not backwards compatible, and it was renamed to `@solana/kit` for version `2.1.0`. This updates the generated code to work with the kit.

### DIFF
--- a/packages/validator-bonds-codama/src/shared/index.ts
+++ b/packages/validator-bonds-codama/src/shared/index.ts
@@ -9,21 +9,21 @@
 import {
   AccountRole,
   isProgramDerivedAddress,
-  isTransactionSigner as web3JsIsTransactionSigner,
+  isTransactionSigner as kitIsTransactionSigner,
   type Address,
   type IAccountMeta,
   type IAccountSignerMeta,
   type ProgramDerivedAddress,
   type TransactionSigner,
   upgradeRoleToSigner,
-} from '@solana/web3.js'
+} from '@solana/kit'
 
 /**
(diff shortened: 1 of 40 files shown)


packages/validator-bonds-codama/src/shared/index.ts
```diff
@@ -9,21 +9,21 @@
...
+} from '@solana/kit'
 
 /**
  * Asserts that the given value is not null or undefined.
  * @internal
  */
 export function expectSome<T>(value: T | null | undefined): T {
-  if (value === undefined || value === null) {
+  if (value === null) {
     throw new Error('Expected a value but received null or undefined.')
```
mwhyte-mrnd, line 27: If value is only checked by `null` now I think you have to update error text too.

mwhyte-mrnd (approved):
👍

packages/validator-bonds-codama/src/shared/index.ts
ochaloup: You are right, as this is a wrong check. The original the codama checks on 'value == null` which considers the `undefined as well`. The typescript linting requires the `===` check, so here is the additional verification to ensure that the undefined is correct.

mwhyte-mrnd (approved):

PR #213, opened 2025-04-16
[pipelines] fixing report type for institutional

ochaloup:
Fixing the Institutional type used for reporting about initialisation and funding settlements in pipeline processing.


ochaloup:
Force-pushed to align with `main` `HEAD`.

martinkrecek (approved):

PR #220, opened 2025-05-05
[GEN-4975] workaround big decimal from institutional payouts

ochaloup:
There happens to be a big decimal numbers or small in scientific notation generated by TS Decimal.
This adds a custom deserializer that manages it and workarounds it.

```
Error: Failure at '--institutional-payouts ./institutional-payouts.json': invalid value: string "4.8076269033743718234e+882", expected a Decimal type representing a fixed-point number at line 11110 column 41
```


ochaloup:
@martinkrecek Thanks for your review and pointing me to my omissions. I updated the check and added a test. WDYT now, please?

martinkrecek (approved):

PR #221, opened 2025-05-09
20250509 fix discord reports [GEN-4999]

kronael:


kronael:
Missed Chalda is OOO...

scripts/generate-discord-public-report.bash
```diff
@@ -100,13 +100,16 @@ do
...
+      reason="BidTooLow"
+    fi
     if [[ $reason_code == "InstitutionalPayout" ]]; then
       reason="Institutional"
     fi
 
     if [[ -z $reason ]]; then
       echo "Unexpected reason code: '$reason_code'" >&2
-      continue
+      reason=$reason_code
```
ochaloup, line 112: I'm unsure about this change. The current behaviour is not letting the unknown code print to control what is publicly shown.  If there is reasoning to print it then ok.

ochaloup (approved):
I understand the reason for

scripts/generate-discord-public-report.bash
kronael: That is a correct reasoning and I'll revert the change.  I was trying to think about how to speed things up by not having a single change require many changes at very different locations.  Can this be done by adding a comment to the definition of reason_code to the original file perhaps?  But I don't see how to do this in a simpler way now, it just ... seems complicated.

kronael, line 112:
```suggestion
      continue
```

ochaloup (approved):

scripts/generate-discord-public-report.bash
ochaloup: The idea of automation makes sense. I completely understand that as the projects aren't linked — it's often unclear which other projects need to be updated when making changes or adding code. I run into the same all the time.

Regarding the automation idea: are you thinking of adding a comment to the original `.rs` file and then parsing both the source and the output JSON? That sounds a bit complicated. In cases like protected events, for example, manual processing might still be necessary because the parsing is more complex.

kronael: "In cases like protected events, for example, manual processing might still be necessary because the parsing is more complex" That is interesting...

With the comment... I meant only having a warning to a potential changing author that this is propagated publically.

This here, however, puts an idea into my head that you could specify the public facing message in the event itself and here it would only get printed.

PR #225, opened 2025-05-19
[fix/trivial] Hotfix penalty calculation

kronael:


ochaloup:
Oh, there is a lint error, but if this will be reverted in a bit I believe it will be fixed then

PR #229, opened 2025-05-23
More info on institutional pipeline [GEN-5100][GEN-4883]

ochaloup:
A way to list the data within the payouts JSON, and whether it is reasonable to process it on-chain. I've got to this processing bash that lists data about payouts. This could need some tuning in the future, but having this as a start seems okayish.
Example listing: https://buildkite.com/marinade-finance/prepare-institutional-distribution/builds/64#0196fc5c-1122-42a8-a4d1-39e9e5b72f85

+ fixing an issue in PSR listing data for settlements in the initial settlement pipeline that shows the wrong info

### DIFF
--- a/.buildkite/prepare-institutional-distribution.yml
+++ b/.buildkite/prepare-institutional-distribution.yml
@@ -85,6 +85,24 @@ steps:
     artifact_paths:
       - "./discord-public-report.txt"
 
+  - label: "📓 Details from payouts data"
+    commands:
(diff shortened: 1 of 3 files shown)


ochaloup:
Taking advantage of not reviewed PR and adding small update (One more script change to better format listing output (I'm not going to do any further changes at this PR, really! 😊 )

martinkrecek (approved):

martinkrecek (approved):

PR #230, opened 2025-05-27
[pipeline] adjust reporting for institutional [GEN-5101]

ochaloup:
Adding a possibility to adjust reporting from the pipeline. Currently for institutional processing the settlement init and fund fails always. That's not what is reasonable as a report that fails everytime makes no sense.

The thing is that payouts usually contain non-institutional validators. It was designed that way as there could be a validator that is removed from the set (by Marinade, intentionally), and we want him to pay the restaking cost. On the other hand, when there are other non-institutional validators, we know that not having a created bond by them is logical, and not being able to create a settlement for deactivating the institutional stake that brought yield to them is ok.

The idea here is to adjust reporting so that it will not fail with non-institutional validators not having a settlement created.

### DIFF
--- a/.buildkite/prepare-institutional-distribution.yml
+++ b/.buildkite/prepare-institutional-distribution.yml
@@ -9,6 +9,7 @@ env:
   CLAIM_TYPE: "institutional-distribution"
   MARINADE_FEE_STAKE_AUTHORITY: 89SrbjbuNyqSqAALKBsKBqMSh463eLvzS4iVWCeArBgB
   MARINADE_FEE_WITHDRAW_AUTHORITY: 89SrbjbuNyqSqAALKBsKBqMSh463eLvzS4iVWCeArBgB
+  INSTITUTIONAL_URL: 'https://institutional-staking.marinade.finance/v1/validators'
 
 steps:
   - label: ":hammer_and_wrench: :rust: Build"
@@ -136,6 +137,7 @@ steps:
               NOTIFY_FEED: $NOTIFY_FEED
               SLACK_FEED: $SLACK_FEED
               NOTIFY_DISCORD_FEED: false
+              INSTITUTIONAL_URL: $INSTITUTIONAL_URL
               gs_bucket: $gs_bucket
       EOF
 
--- a/settlement-pipelines/src/arguments.rs
+++ b/settlement-pipelines/src/arguments.rs
@@ -48,7 +48,7 @@ pub struct GlobalOpts {
 
     /// URL to fetch institutional validators (e.g., https://institutional-staking.marinade.finance/v1/validators).
     /// When set, CLI reporting changes: errors for non-institutional validators are not treated as critical.
-    #[arg(long)]
+    #[arg(long, env)]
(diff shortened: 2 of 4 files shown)


settlement-pipelines/src/reporting.rs
```diff
@@ -98,45 +325,78 @@ impl ErrorHandler {
...
+    }
+
+    pub fn get_retryable_errors(&self) -> Vec<&ErrorEntry> {
+        self.entries.iter().filter(|e| e.is_retryable()).collect()
+    }
+
+    pub fn get_infos(&self) -> Vec<&ErrorEntry> {
+        self.entries
+            .iter()
+            .filter(|e| e.severity() == ErrorSeverity::Info)
```
martinkrecek, line 352: (non-blocking)
It would make sense to me to create `is_info` in `ErrorEntry` (~same implementation like for other severity levels) and use it here, so it would be consistent with other `get_*` functions above.

settlement-pipelines/src/reporting.rs
```diff
@@ -145,14 +405,20 @@ impl ErrorHandler {
...
     report_handler.print_report().await;
     match main_result {
         // when Ok is returned we consult the reality with report handler
-        Ok(_) => CliResult(report_handler.finalize()),
+        Ok(_) => {
+            // before returning result from finalize make possible to adjust entries
+            report_handler
+                .reportable
+                .transform_on_finalize(&mut report_handler.error_handler.entries);
+            CliResult(report_handler.finalize())
```
martinkrecek, line 420: (non-blocking)
There might be some Rust limits around this or it could be impractical for some reasons, but it would feel natural to me to run `self.reportable.transform_on_finalize(&mut self.error_handler.entries)` (or something like that if possible) as the first thing in `report_handler.finalize()` implementation. That way, the caller wouldn't need to think about not forgetting to call `transform_on_finalize` first, it would be done automatically as part of `finalize` when called.

settlement-pipelines/src/reporting.rs
```diff
@@ -45,45 +50,267 @@ impl<T: PrintReportable> DerefMut for ReportHandler<T> {
...
+
+    pub fn with_vote<V: Into<Pubkey>>(mut self, vote_account: V) -> Self {
+        self.vote_account = Some(vote_account.into());
+        self
+    }
+
+    pub fn add(self) {
+        let message = self
+            .message
+            .expect("Message is required - use with_msg() or with()");
```
martinkrecek, line 268:
```suggestion
            .expect("Message is required - use with_msg() or with_err()");
```
No `with()` is present, I guess `with_err()` was meant.

settlement-pipelines/src/reporting.rs
ochaloup: It sounds reasonable. I need to change the code a little bit.

ochaloup:
@martinkrecek the PR updated

ochaloup:
@martinkrecek I'm kindly asking for a re-review. Before the merge, I was re-testing that in the pipeline, and I found some missed parts.

PR #244, opened 2025-06-16
[pipeline] update report script

ochaloup:
Updating report script to show in table amount of validator rewards.

image

### DIFF
--- a/scripts/generate-select-validator-report.bash
+++ b/scripts/generate-select-validator-report.bash
@@ -88,7 +88,7 @@ done < <(<"$json_file" jq -r '.payoutStakers[] | "\(.voteAccount)|\(.deactivatin
 declare -a select_validators
 declare -a non_select_validators
 
-while IFS='|' read -r vote_account apy total_rewards institutional_effective institutional_deactivating commission mev_commission psr_fee_lamports deactivating_payout; do
+while IFS='|' read -r vote_account apy total_rewards validator_rewards institutional_effective institutional_deactivating commission mev_commission psr_fee_lamports deactivating_payout; do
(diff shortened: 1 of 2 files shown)


ochaloup:
@martinkrecek thanks for the review. I wanted to update only the first part of the table and I wrongly added updates on the second non-select one as well. 
I changed the PR. Hopefully, addressing your objections from the comments above.

ochaloup:
I updated the script with a fix to show the PSR penalty data.

ochaloup:
@martinkrecek, as of recent changes in `main` branch, I rebased and forced-push. The source is ready for review (hopefully no more changes here)

scripts/generate-select-validator-report.bash
```diff
@@ -110,56 +126,68 @@ while IFS='|' read -r vote_account apy total_rewards institutional_effective ins
...
+            "$formatted_total_effective" \
+            "$formatted_validator_rewards" \
+            "$formatted_institutional_effective" \
             "$formatted_penalty" \
             "$formatted_staker_payout" \
             "$formatted_distributor_payout"
 
         total_total_rewards=$((total_total_rewards + total_rewards))
-        total_effective=$((total_effective + institutional_effective))
+        total_effective=$((total_effective + total_effective))
```
martinkrecek, line 173: I think this "total variable" will need to be renamed to something like `total_total_effective`, because now the local variable is named the same way as the total one and some unexpected results would be calculated afaik.

martinkrecek (commented):
One issue with variables naming/shadowing found, otherwise looks ok now. 👍

martinkrecek (dismissed):

martinkrecek (approved):

PR #247, opened 2025-06-20
Implement BlacklistPenalty claims [GEN-5371]

kronael:


janlegner (approved):

ochaloup:
@kronael 

I believe this was already clear from the perspective of the failing pipelines, but since this has now happened a second time, I’d like to emphasize that the validator bonds processing relies on JSON data from the previous epoch.
So whenever there’s a change in the DS pipeline processing, that change needs to be reflected in the bonds pipeline starting from `epoch + 1`.

One more thing — though it's not a major concern at the moment — wouldn’t it make sense to consider backward compatibility? For example, making any new field optional would allow us to still process older JSON data when needed (e.g., for investigations). Or maybe even introduce versioning in the JSON format.
Probably no time for this now, but just a thought for making things more seamless in the future.

PR #249, opened 2025-06-22
[trivial] 20250621 hotfix sam meta revert

kronael:
Revert the last hotfix.  To be merged at epoch 808.


ochaloup:
@kronael is this to be closed or should be something checked here (I cannot see any changes in the PR diff)

ochaloup (dismissed):

ochaloup (approved):
Just thinking if the `Optional` could be beneficial for some backwards compatibility purposes - when using the current code with the past data - but np with this update.

kronael:
It can easily be fixed.  I'd opt for simpler code now over backwards compatibility that is not needed.  But I agree it would be purer to just have the backwards compatibility there.  Is there a case where we'd use it?

PR #262, opened 2025-07-21
Close settlement fix plus cli [GEN-5841]

ochaloup:
Some of the settlements were failing to be closed during last week. Investigation shows that the stake accounts used for returning rent were not filtered to those that own enough funded SOLs.

* Adding CLI that helps to solve problematic cases
* CLI helps to investigate on the on-chain failures
* fixing the pipeline to better inform about the reason and to chose to right (better) account


packages/validator-bonds-cli/__tests__/test-validator/closeSettlement.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      rentCollector: rentCollector.publicKey,
+      currentEpoch: expirationEpoch,
+    })
+
+    expect(
+      await provider.connection.getAccountInfo(settlementAccount),
+    ).not.toBeNull()
+    const currentEpoch = (await program.provider.connection.getEpochInfo())
+      .epoch
+    if (expirationEpoch <= currentEpoch) {
```
martinkrecek, line 78: (non-blocking)
It seems to me that if `expirationEpoch` is `0` then this check is always `true`, isn't it? I don't think it breaks the test, it just waits for the next epoch every time (as I understand it), but I'd say that the check is not useful for anything.

Maybe the intention was to have the comparison the other way around and thus wait for the next epoch only when the `currentEpoch` is lower than `expirationEpoch` (which would generally be useful only in cases when `expirationEpoch` is higher than `0`)? I mean, if the check would be `==` and not `<=` then the logic seems right for `expirationEpoch` set to `0`, if the check would be `>=` then the logic would make sense to me for any `expirationEpoch`. In the current state, I think that it works correctly for `expirationEpoch` set to `0`, but it's not entirely correct logic-wise.

martinkrecek (approved):

packages/validator-bonds-cli/__tests__/test-validator/closeSettlement.spec.ts
ochaloup: The `expirationEpoch` is `0` when the test is run solo. When it is run in a group, then the test validator is created only at the start and the epochs are slowly growing. When this test is not the first one; then it is not required to wait for the next epoch to be possible to close the settlement.

As you mentioned, the equation could be rather `===` to be more clear.

PR #264, opened 2025-07-26
Fix ("hotfix") of compute limit budget for tx [GEN-5943]

ochaloup:
The validators report issue on funding bonds with CLI.


packages/validator-bonds-cli-core/src/commands/manage/fundBond.ts
```diff
@@ -49,18 +49,26 @@ export function configureFundBond(program: Command): Command {
         '(default: wallet keypair)',
       parseWalletOrPubkey,
     )
+    .option(
+      '--compute-unit-limit <number>',
+      'Compute unit limit for the transaction (default value based on the operation type)',
```
martinkrecek, line 54:
```suggestion
      'Compute unit limit for the transaction',
```
(non-blocking)
This seems to have a specific default value of `FUND_BOND_LIMIT_UNITS` instead of being based on the operation type, so this might be a copy-paste error.

packages/validator-bonds-cli-core/src/commands/manage/initBond.ts
```diff
@@ -50,6 +50,12 @@ export function configureInitBond(program: Command): Command {
       'Rent payer for the account creation (default: wallet keypair)',
       parseWalletOrPubkey,
     )
+    .option(
+      '--compute-unit-limit <number>',
+      'Compute unit limit for the transaction (default value based on the operation type)',
```
martinkrecek, line 55:
```suggestion
      'Compute unit limit for the transaction',
```
(non-blocking)
This seems to have a specific default value of `INIT_BOND_LIMIT_UNITS` instead of being based on the operation type, so this might be a copy-paste error.

packages/validator-bonds-cli-core/src/commands/manage/initWithdrawRequest.ts
```diff
@@ -72,6 +72,12 @@ export function configureInitWithdrawRequest(program: Command): Command {
       'Rent payer for the account creation (default: wallet keypair)',
       parseWalletOrPubkey,
     )
+    .option(
+      '--compute-unit-limit <number>',
+      'Compute unit limit for the transaction (default value based on the operation type)',
```
martinkrecek, line 77:
```suggestion
      'Compute unit limit for the transaction',
```
(non-blocking)
This seems to have a specific default value of `INIT_WITHDRAW_REQUEST_LIMIT_UNITS` instead of being based on the operation type, so this might be a copy-paste error.

packages/validator-bonds-cli-core/src/commands/manage/mintBond.ts
```diff
@@ -33,18 +33,26 @@ export function configureMintBond(program: Command) {
       'Rent payer for the mint token account creation (default: wallet keypair)',
       parseWalletOrPubkey,
     )
+    .option(
+      '--compute-unit-limit <number>',
+      'Compute unit limit for the transaction (default value based on the operation type)',
```
martinkrecek, line 38:
```suggestion
      'Compute unit limit for the transaction',
```
(non-blocking)
This seems to have a specific default value of `MINT_BOND_LIMIT_UNITS` instead of being based on the operation type, so this might be a copy-paste error.

martinkrecek (approved):
Seeing the scope of changes required for this, I'm thinking that going forward it might make sense to add some common params/params to be shared among the individual commands if a similar change would occur again. But no high prio, I would not rework it in this PR, just sharing the thoughts.

ochaloup:
As the issue is of copy&paste error, I consider some abstraction on option creation is needed. I think having the option creation as a special function is all good enough here.

martinkrecek (approved):

PR #268, opened 2025-09-02
CLI: permit validator identity to be used when showing bond info [GEN-6189]

ochaloup:
Making it possible to use the validator identity to show the bond account

### DIFF
--- a/packages/validator-bonds-cli-core/src/utils.ts
+++ b/packages/validator-bonds-cli-core/src/utils.ts
@@ -23,11 +23,13 @@ import {
   PublicKey,
   SendTransactionError,
   StakeProgram,
+  SystemProgram,
 } from '@solana/web3.js'
 import { Logger } from 'pino'
 import { setProgramIdByOwner } from './context'
 import BN from 'bn.js'
 import { logDebug } from '@marinade.finance/ts-common'
+import { findVoteAccountByIdentity } from '@marinade.finance/validator-bonds-sdk'
 
 /**
  * Expecting the provided address is a bond or vote account,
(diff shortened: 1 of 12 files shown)


packages/validator-bonds-cli-core/src/utils.ts
```diff
@@ -44,23 +46,45 @@ export async function getBondFromAddress({
...
-  let voteAccountAddress = await isVoteAccount({
-    address,
-    accountInfo,
-    logger,
-  })
+  // Let's consider this could be a validator identity account
+  let voteAccountAddress: PublicKey | null = null
+  if (
+    accountInfo === null || // no account exists at the address
+    (accountInfo !== null && accountInfo.owner.equals(SystemProgram.programId))
```
mwhyte-mrnd, line 62: I think you can just:
```ts
if (accountInfo === null || accountInfo.owner.equals(SystemProgram.programId))
```

packages/validator-bonds-sdk/src/web3.js/voteAccount.ts
```diff
@@ -18,3 +32,41 @@ export async function getRentExemptVote(
...
+  connection,
+  identity,
+  logger,
+}: {
+  connection: Provider | Connection | HasProvider
+  identity: PublicKey
+  logger?: LoggerPlaceholder
+}): Promise<ProgramAccountInfo<Buffer> | undefined> {
+  const filters: GetProgramAccountsFilter[] = []
+  filters.push({
```
mwhyte-mrnd, line 46: Why it's not defined straight on line 45?

mwhyte-mrnd (dismissed):

packages/validator-bonds-cli-core/src/utils.ts
ochaloup: Ah, yes, sure, crap, thanks.

packages/validator-bonds-sdk/src/web3.js/voteAccount.ts
ochaloup: Yes, it makes sense. It is a bit leftover from the development when I was thinking of adding one more filter and deciding based on the existence of an optional function parameter. Thanks, fixing.

PR #282, opened 2025-10-03
Bid pipeline sanity check [GEN-6071]

ochaloup:
This is "ready to review," but it needs two prior PRs to be checked/reviewed.

* https://github.com/marinade-finance/institutional-staking/pull/61
* https://github.com/marinade-finance/validator-bonds/pull/281

(+ maybe to take the function from ts-common https://github.com/marinade-finance/typescript-common/pull/41)


ochaloup:
@kronael The PR introduces the semi-automatic pipeline processing for bonds. It is ready to review. I am still struggling with how to set up the data points correctly. But I would like to have this merged and then add/tune the parameters.

packages/validator-bonds-sanity-check/src/commands/check.ts
```diff
@@ -0,0 +1,132 @@
...
+        scoreThreshold,
+        type,
+      }: {
+        current: string
+        past: string[]
+        correlationThreshold: Decimal
+        scoreThreshold: Decimal
+        type: ProcessingType
+      }) => {
+        await manageCheck({
```
kronael: again.  if the names matched,  could it be just done as `action(manageCheck)`?

packages/validator-bonds-sanity-check/src/context.ts
```diff
@@ -0,0 +1,28 @@
...
+
+  static define({
+    logger,
+    commandName,
+  }: CLIContextConfig): SanityCheckCLIContext {
+    const context = new CLIContext({
+      logger,
+      commandName,
+    })
+    setContext(context)
```
kronael: This is a bit ugly.  But I guess passing it around as ctx - first var is tedious or impossible with Comander.

packages/validator-bonds-sanity-check/src/anomalyDetector.ts
```diff
@@ -0,0 +1,232 @@
...
+  historicalSettlements: SettlementsDto[]
+  type: ProcessingType
+  // z-score threshold (e.g., 2.0 means 2 standard deviations from mean)
+  scoreThreshold: Decimal
+  // talking in percent ratio, e.g. 0.15 = 15%
+  correlationThreshold: Decimal
+  logger?: LoggerPlaceholder
+}): StatsCalculation[] {
+  if (historicalSettlements.length < 3) {
+    logWarn(
```
kronael: This does not throw?

packages/validator-bonds-sanity-check/src/anomalyDetector.ts
```diff
@@ -0,0 +1,232 @@
...
+      logger,
+      'Not enough historical data for reliable anomaly detection, please provide at least 3 epochs.',
+    )
+  }
+
+  const currentData = transform(currentSettlements)
+  const historicalData = historicalSettlements.map(transform)
```
kronael: Love how the function fits its use perfectly as opposed to having a closure envelope around at every call site.

packages/validator-bonds-sanity-check/src/anomalyDetector.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+
+  return calculations
+}
+
+export function reportAnomalies({
+  currentSettlements,
+  historicalSettlements,
+  type,
+  scoreThreshold = new Decimal(2.0), // working with z-score like threshold
```
kronael: Should be "2.0" and "0.15" to be "correct.  Imho.

packages/validator-bonds-sanity-check/src/anomalyDetector.ts
ochaloup: I was not sure and I decided to leave it for warn. From your point I believe you consider it should. Then I'm changing it.

ochaloup:
@kronael, thanks a lot for the review points! I addressed most of them. Asking for re-review.

packages/validator-bonds-sanity-check/src/commands/checkSettlement.ts
```diff
@@ -0,0 +1,109 @@
...
+    merkleTreesPath,
+  )
+
+  logger.debug(
+    `Loaded settlements from ${settlementsPath} and merkle trees from ${merkleTreesPath}`,
+  )
+
+  // Check 0: Epoch comparison
+  if (settlements.epoch !== merkleTrees.epoch) {
+    throw CliCommandError.instance(
```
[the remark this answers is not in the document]
kronael: Makes sense.  I obviously did not understand the whole picture.  Sorry.

packages/validator-bonds-sanity-check/src/context.ts
ochaloup: Yes, it is a global variable. For the case of the context it seems to me as a reasonable usage that is straightforward. I would not be using it for normal parameter passing though.

PR #388, opened 2026-04-09
[feat] charge activating stake in bid settlements per new formula [GEN-7606]

kronael:
Introduces activating stake charging in bid settlements with a proportional fee split.                                                                                                       
  Activating stakers are separated into a dedicated PriorityFee settlement,                                                                                                                    
  and SettlementClaim.stake_amount is split into active_stake + activating_stake (always present in JSON).
                                                                                                                                                                                               
  Updates the TypeScript DTO, Discord report, and all test fixtures to match the new schema.                                                                                                   
  Backwards compat preserved via serde rename on active_stake.


kronael:
The active > 0 is what Claude suggested mostly.  I'll need to check it once more today, but IMHO its good for review.

ochaloup (commented):
This seems to me ok. Few comments though

* Activating charge bypasses distributor fee `total_marinade_stakers_rewards` doesn't include activating_bid_claim, so the Marinade/DAO fee is never applied to the activating charge. Is it by design? If so it would be good to comment on the intent.
* If all marinade accounts for a validator are brand-new (`active_delegation_lamports == 0`), then `total_marinade_active_stake == 0` and the validator is skipped. Any `total_marinade_activating_stake` accumulated is silently discarded. Maybe to add some log for it to know?
* Maybe enhance the testing with case of having two accounts - one fully active (so the validator isn't skipped) and one brand-new (`active = 0, activating > 0`)

kronael:
Thank you those are very good comments and pointing ofc towards gross bugs.  Will fix and resubmit.

ochaloup (commented):
When looking at the code, the method was already long and is now even longer. Probably not necessary right now, but we should consider splitting it into more readable parts that reflect the step-by-step functionality.
Otherwise, looks good to me.

Then I worked with Claude to help me localize issues, and these seem (at least partially) relevant to me.

---
---

1.
`SettlementClaim` documents `stake_accounts` as `"pubkey -> active_delegation_lamports"` (`settlement_collection.rs:18`) and the field is named `active_stake`. For activating claims, the code populates these with `activating_delegation_lamports` instead:

```rust
// bidding.rs — activating stakers section
.map(|s| (s.pubkey, s.activating_delegation_lamports))  // not active_delegation_lamports
...
active_stake: activating_sum,  // actually activating lamports
```

**Impact:**
- Breaks the documented contract of `SettlementClaim`.
- Settlement JSON files (used for auditing) will show activating lamports under a field called `active_stake` and in a map documented as `active_delegation_lamports`.
- When `merge_claims()` in `merkle_generator.rs` merges an active claim and an activating claim for the same `(withdraw_authority, stake_authority)`, the resulting `active_stake` becomes `sum(active_delegation) + sum(activating_delegation)` — a semantically meaningless number.

**Mitigation:** This does NOT affect on-chain correctness — the merkle tree only uses `claim_amount`, not `active_stake` or `stake_accounts` values. But it degrades audit quality.

**Suggestion:** Either (a) rename or add an explicit field for the lamport type (e.g. `stake_lamports` instead of `active_stake`), or (b) add a comment at the activating claim construction clarifying that `active_stake` is overloaded here and why it's safe.

---

2.
```
The test only asserts `claims_amount == 200_000_000`. It does not verify:
. That there are exactly 2 claims.
. That the active staker (stake_account(1)) receives 100M.
. That the activating staker (stake_account(2)) receives 100M.
. There are claims for the activating staker (stake_account(2)).
. The active staker (stake_account(1)) gets nothing or a separate claim depending on whether static_bid was applied (it's 0 in this test, so no active claim expected).
```

kronael:
I am thinking about these, and its not easy...  I'd just split the settlements into different Settlements (the new one with PriorityFee) and change the meaning of the stake in SettlementClaim.. .but am not sure where everywhere its used.   Definitely the merging would stop making sense, but if its only for debugging, maybe the debugging infos do not need to be merged?

settlement-distributions/bid-distribution/src/generators/psr_events.rs
```diff
@@ -82,7 +76,7 @@ fn generate_psr_settlements_for_config(
                     .iter()
                     .map(|s| (s.pubkey, s.active_delegation_lamports))
                     .collect();
-                let active_stake = stake_accounts.values().sum();
+                let active_stake: u64 = stake_accounts.values().sum();
```
kronael:
```suggestion
                let active_stake = stake_accounts.values().sum();
```

kronael:
Did reshuffle how the "tracing" works.  Will be testing the main logic still, but I kind of trust the LLM with tests if I say to focus on them...

settlement-distributions/merkle-generator/src/merkle_generator.rs
```diff
@@ -85,40 +78,19 @@ fn validate_sources(sources: &[SettlementSource]) -> anyhow::Result<(u64, u64)>
...
+    let mut amount_map: HashMap<SettlementKey, u64> = HashMap::new();
 
     for claim in claims {
-        let key = ClaimKey {
+        let key = SettlementKey {
             withdraw_authority: claim.withdraw_authority,
             stake_authority: claim.stake_authority,
         };
-
-        claim_map
```
ochaloup, line 101: 🤟

settlement-distributions/settlement-common/src/settlement_collection.rs
```diff
@@ -15,17 +15,24 @@ pub struct SettlementClaim {
     pub withdraw_authority: Pubkey,
     #[serde(with = "pubkey_string_conversion")]
     pub stake_authority: Pubkey,
-    /// stake account pubkey -> active_delegation_lamports
+    /// stake account pubkey -> stake lamports (active or activating depending on settlement type)
     #[serde(with = "map_pubkey_string_conversion")]
     pub stake_accounts: HashMap<Pubkey, u64>,
-    pub active_stake: u64,
```
ochaloup, line 21: If there is no really strong reason to rename this, please stick with the long-term used `active_stake`. It may not be descriptive enough, but we have been using it in JSON processing for a long time, and breaking compatibility here with some TS processing (which can be adjusted in the repo), as well as with older JSONs and/or data processing that works with data from GCloud, does not make sense to me.

ochaloup (changes requested):
Please, let’s stay backward compatible with names that are serialized into JSON.

settlement-distributions/settlement-common/src/settlement_collection.rs
```diff
@@ -15,17 +15,27 @@ pub struct SettlementClaim {
...
+    // Renamed to stake_amount in Rust (holds active *or* activating lamports depending on
+    // settlement type), but kept as "active_stake" in JSON for backwards compatibility.
+    #[serde(rename = "active_stake")]
```
ochaloup: I'm not fond of this solution; the Rust code differs from the JSON output, and it seems to me it can be misleading. But I understand that you're changing the semantics of the field, so you're trying to align those. It's a trade-off, and since I'm pushing for backward compatibility, I would push it here as well. But whatsoever.

ochaloup (dismissed):

kronael:
image

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -318,92 +365,165 @@ pub fn generate_bid_settlements(
...
-                    settlement_claim_sum,
-                    validator.vote_account
-                );
+                bidding_claims_amount += dao_fee_for_bidding;
+            }
+            if marinade_fee_for_priority > 0 {
+                priority_fee_claims.push(SettlementClaim {
+                    withdraw_authority: authorities.marinade_withdraw,
+                    stake_authority: authorities.marinade_stake,
+                    stake_accounts: marinade_fee_deposit_stake_accounts.clone(),
```
ochaloup: This field is more about reporting than actual usage, I think, but
`marinade_fee_deposit_stake_accounts` and `dao_fee_deposit_stake_accounts` come from `get_fee_deposit_stake_accounts` (`generators/mod.rs`), which maps `stake.pubkey -> active_delegation_lamports`. Here, it should probably include those that have some activating stake, I think?

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -318,92 +365,165 @@ pub fn generate_bid_settlements(
...
 
-            if marinade_fee_claim > 0 {
-                let authorities = fee_config.fee_authorities();
-                claims.push(SettlementClaim {
+            // Split fee claims proportionally between active and activating pools
+            let activating_fee_fraction = if stakers_total_claim > 0 {
+                Decimal::from(activating_stakers_pool) / Decimal::from(stakers_total_claim)
+            } else {
+                Decimal::ZERO
+            };
```
ochaloup, line 457: An edge case found by claude.
If the distributor fee percentage equals 100% then `stakers_total_claim == 0`, `activating_fee_fraction == 0`, and 100% of both fees flow to the `Bidding` settlement - even if the entire gross charge came from activating stake.
The PriorityFee settlement then has no claims at all and is not emitted, yet the activating-origin fee lamports sit in a Bidding settlement.
The lamports are still correctly collected as a fee claim, but they end up in the wrong settlement type

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -416,15 +536,37 @@ pub fn generate_bid_settlements(
...
             };
             let details_json = serde_json::to_value(&settlement_details)?;
 
+            let priority_fee_details = PriorityFeeSettlementDetails {
+                total_marinade_activating_stake,
+                total_marinade_uncharged_activating_stake,
+                activating_stake_pmpe: validator
+                    .rev_share
+                    .activating_stake_pmpe
+                    .unwrap_or(Decimal::ZERO)
```
ochaloup, line 548: A note/suggestion: `activating_stake_pmpe = None` is indistinguishable from `Some(Decimal::ZERO)` in details
A suggestion: Either skip writing the priority-fee settlement entirely when `activating_stake_pmpe` is `None` and `total_marinade_activating_stake == 0`, or make the field `Option<String>`

ochaloup (commented):
It seems correct to me, though I have some doubts about the stake accounts emitted for the activating stake.

(a note for myself: I need to mention that once this is merged, I need to refactor `bidding.rs` a bit. It is really huge and hard to understand.)

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: That would be definitely better.  I will review this once more.  There were some thinking behind that, but I apparently missed something.

kronael: The idea definitely is : ONLY stake accounts WITH activating stake RECEIVE the priority bid.  Not anyone else.  And those should also be the ones reported, I guess.  At least that would be the intention, but I did not check the marinade claims explicitly.  note to self : Cluade just write bugs on every line ... pretty awesome.

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: I'd just skip that.  I don't like None.

ochaloup: Hm... I like :-)

kronael: Actually.  Here None mean its just not in ds-sam, which means its an old auction.  This makes me think that the 0 is actually correct and None would not be.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -315,95 +359,169 @@ pub fn generate_bid_settlements(
...
                 stakers_total_claim,
                 validator.vote_account
             );
 
-            if marinade_fee_claim > 0 {
-                let authorities = fee_config.fee_authorities();
-                claims.push(SettlementClaim {
+            // Split fee claims proportionally between active and activating pools
+            let activating_fee_fraction = if stakers_total_claim > 0 {
+                Decimal::from(activating_stakers_pool) / Decimal::from(stakers_total_claim)
```
ochaloup, line 427: Minor: `activating_stakers_pool` was derived at line 342 as
`(Decimal::from(stakers_total_claim) * activating_fraction).to_u64().unwrap_or(0)`.
So `activating_stakers_pool / stakers_total_claim ≈ activating_fraction` but diverges
by the u64-truncation loss (up to ~1/`stakers_total_claim`). 

The `activating_stakers_pool` could be of type `Decimal`

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -315,95 +359,169 @@ pub fn generate_bid_settlements(
...
-                    settlement_claim_sum,
-                    validator.vote_account
-                );
+                bidding_claims_amount += dao_fee_for_bidding;
             }
+            if marinade_fee_for_priority > 0 {
+                priority_fee_claims.push(SettlementClaim {
+                    withdraw_authority: authorities.marinade_withdraw,
+                    stake_authority: authorities.marinade_stake,
+                    stake_accounts: fee_deposit.marinade_activating.clone(),
```
ochaloup: NOTE: same pattern for dao_fee_for_priority at lines 503-513

Consider listing to `marinade_active` stake accounts instead of activating.

All other penalties list the stake accounts where the DAO penalty will be assigned.

```
// Bidding Marinade fee — bidding.rs:470-480
stake_accounts: fee_deposit.marinade_active.clone(),   // <- the single Marinade fee-deposit account only

// Bidding DAO fee — bidding.rs:481-491
stake_accounts: fee_deposit.dao_active.clone(),        // <- the single DAO fee-deposit account only
...
```

As this is the one (or a limited number) of stake accounts, we cannot expect any activating stake accounts to exist. Most of the time it will be an empty list of stake accounts, and `[]` will be reported here.


Trade-off acknowledged: within a PriorityFee settlement, the Marinade/DAO fee claim will then carry active_stake > 0 while the staker claims carry activating_stake > 0. That mixed-state-within-one-settlement looks semantically odd, but it's the same asymmetry

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -315,95 +359,169 @@ pub fn generate_bid_settlements(
                 validator.vote_account
             );
 
-            let (marinade_fee_deposit_stake_accounts, dao_fee_deposit_stake_accounts) =
-                get_fee_deposit_stake_accounts(stake_meta_index, fee_config);
+            let fee_deposit = get_fee_deposit_stake_accounts(stake_meta_index, fee_config);
```
ochaloup: Probably not needed to be called on every cycle.

```
let fee_deposit = get_fee_deposit_stake_accounts(stake_meta_index, fee_config);
for validator in sam_validator_metas { ... }
```

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -315,95 +359,169 @@ pub fn generate_bid_settlements(
...
+                });
+                priority_fee_claims_amount += dao_fee_for_priority;
+            }
+            ensure!(
+                bidding_claims_amount + priority_fee_claims_amount <= settlement_claim_sum,
+                "The sum of total claims {} exceeds the total claim amount {} after adding fees for validator {}",
+                bidding_claims_amount + priority_fee_claims_amount,
+                settlement_claim_sum,
+                validator.vote_account
+            );
```
ochaloup, line 502: From a functionality perspective this seems the same, but the previous two separate `ensure!` calls may provide more visible reasoning.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -300,6 +332,18 @@ pub fn generate_bid_settlements(
             // minimum is 0 when distributor fee is of amount of total (stakers get nothing)
             let settlement_claim_sum = settlement_claim.sum_u64()?;
             let stakers_total_claim = settlement_claim_sum.saturating_sub(distributor_fee_claim);
+            // Split stakers_total_claim between active stakers (earned rewards + static bid)
+            // and activating stakers (activating charge), proportional to each pool's share.
+            let activating_fraction = if settlement_claim_sum > 0 {
```
ochaloup: minor: `settlement_claim_sum: u64` is the truncated cast of `settlement_claim.sum()`
maybe `settlement_claim.sum() > Decimal::ZERO` ?

settlement-distributions/bid-distribution/src/generators/mod.rs
```diff
@@ -15,43 +15,60 @@ use std::collections::HashMap;
...
 
-/// The output Settlements data is updated with stake accounts owned by Marinade and DAO
+pub struct FeeDepositStakeAccounts {
+    pub marinade_active: HashMap<Pubkey, u64>,
+    pub marinade_activating: HashMap<Pubkey, u64>,
+    pub dao_active: HashMap<Pubkey, u64>,
+    pub dao_activating: HashMap<Pubkey, u64>,
+}
+
 pub fn get_fee_deposit_stake_accounts(
```
ochaloup, line 27: The comment `The output Settlements data is updated with stake accounts owned by Marinade and DAO` was removed. Maybe some similar note should be left on the `FeeDepositStakeAccounts`?
(` /// Stake accounts owned by Marinade and DAO that fee claims in the output settlements are routed to.`)

ochaloup (commented):
I did verification of how the code runs against processed epoch and that all seems fine. I updated a test script to be in the repository (now or later), https://github.com/marinade-finance/validator-bonds/tree/20260409_pay_for_priority-script

For the review:

a) It would be good to update the ds-sam sdk version used here to 0.0.48

b) I'm discussing the processing with Claude and we got to this. Can you confirm if this line of thought is reasonable, or if there's a flaw in some basic assumption?

```
 total_marinade_uncharged_activating_stake is accumulated at bidding.rs:125-126, inside the skip branch of the same if meta.active_delegation_lamports == 0 { ... } else { ... } that finding #2 argues against.

  The exact source:

  // bidding.rs:121-126
  if meta.active_delegation_lamports == 0 {
      total_marinade_activating_stake +=
          meta.activating_delegation_lamports;
  } else {
      total_marinade_uncharged_activating_stake +=
          meta.activating_delegation_lamports;  // <── here
      ...
  }

Running sum per validator, per epoch, of meta.activating_delegation_lamports for every stake account that fell into the skip branch (active > 0 && activating > 0).

What's wrong: the field is only surfaced in two places — an info! log and PriorityFeeSettlementDetails.total_marinade_uncharged_activating_stake. The latter is serialized into the settlement's details, but add_to_settlement_collection at mod.rs:81 drops the whole PriorityFee settlement when priority_fee_claims.is_empty(). So in the worst case — a validator whose activating stake entirely falls into the skip branch — the counter never reaches downstream reporting. The one scenario where this diagnostic matters most is the one where it disappears.

How to fix:
 - If the skip branch is removed - this works because in this system an activating lamport can't persist across epochs, so there's no prior-epoch charge to double up against. Delete it along with the skip logic.
- If the skip is kept as deliberate top-up policy: always emit the diagnostic when total_marinade_uncharged_activating_stake > 0, regardless of priority_fee_claims.
```

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: This is just some auto-simplification, so probably I shoudnt've touched these...

kronael:
b) True that if some lamports are activated, the failsafe monitoring is not really needed.  Would just remove it then

ochaloup:
@kronael what do you think about the discussion about the activating lamports above https://github.com/marinade-finance/validator-bonds/pull/388#pullrequestreview-4139887323

I think the fix within the semantics mentioned would be:

```
// Instead of:
if meta.activating_delegation_lamports > 0 && meta.active_delegation_lamports == 0 {
    total_marinade_activating_stake += meta.activating_delegation_lamports;
}
// Just:
total_marinade_activating_stake += meta.activating_delegation_lamports;
```

Snapshots are taken at the last slot of each epoch, and Solana auto-promotes activating → active at the epoch boundary, so a given activation appears as activating > 0 in exactly one snapshot. Under those semantics the skip branch (active > 0 && activating > 0) can't be mid-warmup — it fires on top-ups (fresh delegation added to an already-active account) and merges (new activating account merged into an active one). Both are genuinely new activating exposure that ds-sam already charges activatingStakePmpe on.

Should the skip be removed, or is it deliberate policy to not charge top-ups?

ochaloup (approved):

PR #393, opened 2026-04-17
[feat] Dynamic SSI-based fee adjustment for bid settlements [GEN-7676]

kronael:
Introduces SSI-based dynamic fee adjustment: the distributor fee is capped                                                                                                                   
  so staker yield never drops below SSI + apy_over_ssi_pmpe (set to -10 to disable).                                                                                                           
  min_fee_bps provides a hard floor that can override the cap.                                                                                                                                                                                                                                            
                                                                                                                                                                                               
  --validator-meta-collection is now required for all SAM settlements since
  SSI computation needs it. Renames: FeePercentages→FeeRates,                                                                                                                                  
  ssi_underperformance_pmpe→apy_over_ssi_pmpe.     

There will be a change in ds-sam adding the `ssi_pmpe`.  We should be able to merge without it and there should be no change whatsoever to the current workings.


packages/bonds-eventing/src/evaluate-deltas.ts
```diff
@@ -64,6 +66,33 @@ function getCapConstraint(v: AuctionValidator): string | null {
...
+function computePenalties(v: AuctionValidator): {
+  total: number
+  bidTooLow: number
+  blacklist: number
+  bondRiskFee: number
+} {
+  const stake = v.marinadeActivatedStakeSol
+  const bidTooLow = (v.revShare.bidTooLowPenaltyPmpe / 1000) * stake
+  const blacklist = (v.revShare.blacklistPenaltyPmpe / 1000) * stake
+  const bondRiskFee = v.values?.bondRiskFeeSol ?? 0
```
ochaloup: These conversions seem to be a good mess. This is the logic of ds-sam duplicated here, related to my wish not to compute anything connected to ds-sam here.

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -49,6 +51,12 @@ pub struct FeeConfig {
     pub marinade_fee_bps: u64,
     pub marinade: AuthorityConfig,
     pub dao: DaoConfig,
+    /// Minimum fee floor in basis points; fee is never reduced below this (default: 0)
+    #[serde(default)]
+    pub min_fee_bps: u64,
+    /// target = SSI + apy_over_ssi_pmpe. Set to -10 (or lower) to effectively disable fee cap.
```
ochaloup: The comment states that `-10` disables the cap, but the default is `0`. Shouldn’t the default be `-10` to have it disabled by default? See B) in my PR review comment.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -321,9 +322,27 @@ pub fn generate_bid_settlements(
...
-                total_marinade_stakers_rewards * fee_percentages.marinade_distributor_fee;
+            let effective_fee = if total_marinade_stakers_rewards > Decimal::ZERO
+                && total_marinade_active_stake > 0
+            {
+                let target = ssi_pmpe + fee_config.apy_over_ssi_pmpe;
+                let staker_yield_pmpe = total_marinade_stakers_rewards
+                    / Decimal::from(total_marinade_active_stake)
+                    * Decimal::ONE_THOUSAND;
+                let fee_cap = (Decimal::ONE - target / staker_yield_pmpe).max(Decimal::ZERO);
+                fee_percentages.marinade_distributor_fee.min(fee_cap).max(fee_percentages.min_fee)
```
ochaloup: Fee `min`/`max` composition can return a fee > 1.0

> `FeeConfig::validate()` caps each of `marinade_fee_bps`, `dao.fee_split_share_bps`, `min_fee_bps` at 10_000, but does not require `min_fee_bps <= marinade_fee_bps`. A misconfiguration (e.g. `marinade_fee_bps = 1000`, `min_fee_bps = 5000`) lets `effective_fee > marinade_distributor_fee` and therefore `minimum_distributor_fee_claim > settlement_claim.sum()`. The `.min(settlement_claim.sum())` clamp on line 346-348 prevents overclaim, so this is a misconfiguration rather than a correctness bug, but the validator name is kept silent.

settlement-distributions/bid-distribution/src/generators/tests.rs
```diff
@@ -2520,3 +2546,120 @@ fn test_psr_epoch_mismatch_returns_error() {
...
+        6 * solana_sdk::native_token::LAMPORTS_PER_SOL,
+        "configured fee (0.30) must apply when fee_cap (0.75) is not binding"
+    );
+}
+
+#[test]
+fn test_ssi_fee_cap_zero_when_ssi_exceeds_yield() {
+    // staker_yield_pmpe=20, ssi=25 → fee_cap=0, min_fee=0 → effective=0 → no fee
+    let settlements = run_ssi_test(25.0, ssi_fee_config(3000, 0, 0.0));
+    let marinade_fee = sum_claims_for_authority(&settlements, &TEST_PUBKEY_MARINADE, &TEST_PUBKEY_MARINADE);
```
ochaloup:
```suggestion
```

settlement-distributions/bid-distribution/src/bin/cli.rs
```diff
@@ -162,6 +159,12 @@ fn main() -> anyhow::Result<()> {
             rewards_collection.total_rewards()
         );
 
+        info!("Computing SSI from validator meta collection...");
+        let validator_meta: ValidatorMetaCollection = read_from_json_file(validator_meta_path)
```
ochaloup: The `ValidatorMetaCollection` is read and deserialized a second time here. Deserialization should probably be performed only once for the CLI.

settlement-distributions/bid-distribution/src/generators/bidding.rs
ochaloup: Not directly related to this PR, but to the previous one.
It would be better not to use `get_fee_deposit_stake_accounts` in a loop, as it relies on the `iter().find()` pattern, which is not performance-friendly.
Some other way to search through the stake data should be considered.

settlement-distributions/bid-distribution/src/ssi.rs
```diff
@@ -0,0 +1,107 @@
...
+/// `validator_meta.validator_rewards` = theoretical 0%-commission inflation (network-wide).
+/// `block_rewards` summed from rewards_collection = actual block rewards (network-wide via
+/// validators_blocks.json).
+pub fn calculate_ssi_pmpe(
+    rewards: &RewardsCollection,
+    validator_meta: &ValidatorMetaCollection,
+) -> anyhow::Result<Decimal> {
+    let total_stake = validator_meta.total_stake();
+    anyhow::ensure!(total_stake > 0, "total stake is zero, cannot compute SSI");
+    let total_block_rewards: u64 = rewards
```
ochaloup: There should be a similar check as above to ensure that `total_block_rewards` is > 0.

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -125,17 +139,6 @@ impl SettlementConfig {
...
-        matches!(self, SettlementConfig::Sam(_))
-    }
-
-    /// Checks if this is a PSR settlement type (DowntimeRevenueImpact, CommissionSamIncrease)
-    pub fn is_psr_settlement(&self) -> bool {
-        matches!(self, SettlementConfig::Psr(_))
-    }
-
-    /// Returns the PSR config reference if this is a PSR settlement type.
     pub fn to_psr_config(&self) -> Option<&PsrSettlementConfig> {
```
ochaloup, line 162: Maybe add `to_sam_config()` to help deduplicate the code and show how to work when a new SAM kind is added.

ochaloup (changes requested):
A) I don't like this approach from a data flow point of view.

I would prefer PMPE not to be computed here in the distribution phase. This code was meant to stay very small—just fetch data and recompute claims. Now it's turning into a heavy data platform. Can we instead do it like this:

> ds-sam emits `ssi_pmpe` as part of its auction output, alongside `totalPmpe`, etc. The bonds repo just reads it. That decouples the methodology from the settlement pipeline and makes the cap a pure applicator rather than a co-implementer of SSI.

B) I'm lost in the arguments here.

* What is the meaning of `apy_over_ssi_pmpe`? Suggesting `staker_yield_target_over_ssi_pmpe` - longer, but more descriptive.
* `min_fee_bps` is not clearly differentiated from `marinade_fee_bps`, and its meaning is unclear. Maybe pair `marinade_fee_bps` (ceiling) with `marinade_fee_floor_bps` (floor). At the very least, improve the CLI help text (they are interconnected). And we should validate `marinade_fee_floor_bps <= marinade_fee_bps`.
* What is that `-10` constant? Please no. That's too magical for me. Let's use `Option<Decimal>` with clear meaning: `None` = cap off, `Some(0)` = target SSI exactly, `Some(x)` = target SSI + x.

C) The Discord report script does not handle the new `PriorityFee` reason.

packages/bonds-eventing/src/evaluate-deltas.ts
kronael: Yes, that is a very good comment and I agree with that and will do it like that.  That is much better.

settlement-distributions/bid-distribution/src/settlement_config.rs
kronael:
```suggestion
    /// target = SSI + apy_over_ssi_pmpe.
```

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -49,6 +51,12 @@ pub struct FeeConfig {
     pub marinade_fee_bps: u64,
     pub marinade: AuthorityConfig,
     pub dao: DaoConfig,
+    /// Minimum fee floor in basis points; fee is never reduced below this (default: 0)
```
kronael:
```suggestion
    /// Minimum fee floor in basis points; fee is never reduced below this
```

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -46,24 +48,39 @@ pub struct DaoConfig {
...
 #[derive(Clone, Deserialize, Serialize, Debug)]
 pub struct FeeConfig {
-    pub marinade_fee_bps: u64,
+    pub max_fee_bps: u64,
     pub marinade: AuthorityConfig,
     pub dao: DaoConfig,
+    #[serde(default)]
+    pub min_fee_bps: u64,
+    #[serde(default)]
+    pub min_yield_premium_over_ssi_pmpe: Decimal,
```
kronael, line 57: We can support disabling once its needed.

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -46,24 +48,39 @@ pub struct DaoConfig {
 /// Shared fee configuration for SAM settlement types (Bidding, BidTooLowPenalty)
 #[derive(Clone, Deserialize, Serialize, Debug)]
 pub struct FeeConfig {
-    pub marinade_fee_bps: u64,
+    pub max_fee_bps: u64,
     pub marinade: AuthorityConfig,
     pub dao: DaoConfig,
+    #[serde(default)]
+    pub min_fee_bps: u64,
```
kronael, line 56: Min fee = 0 is ok.

kronael:
Rebased and now ready for review.

settlement-config.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 fee_config:
-  marinade_fee_bps: 800
+  max_fee_bps: 800
```
kronael, line 4:
```suggestion
  min_fee_bps: 200
  max_fee_bps: 800
```

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -46,24 +48,40 @@ pub struct DaoConfig {
...
+    #[serde(default)]
+    pub min_yield_premium_over_ssi_pmpe: Decimal,
 }
 
 impl FeeConfig {
     /// Validates that fee basis points are within valid range (0..=10000)
     pub fn validate(&self) -> anyhow::Result<()> {
```
ochaloup, line 62: maybe to add `!ensure` for `min_yield_premium_over_ssi_pmpe` that is bigger or equal to 0?

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -10,10 +10,12 @@ use solana_sdk::pubkey::Pubkey;
 /// Fee percentages calculated from basis points
 #[derive(Debug, Clone, Copy, Default)]
```
ochaloup: Thinking whether we should enforce the values to be explicitly defined in `yaml` and drop the `Default` here.

ochaloup (dismissed):
Overall, it looks nice to me, and I really appreciate the separation into ds-sam (https://github.com/marinade-finance/ds-sam/pull/105).

settlement-distributions/bid-distribution/src/settlement_config.rs
kronael: I guess I wanted to be backwards compatible.  But I am all for rather having things defined!

settlement-distributions/bid-distribution/src/settlement_config.rs
kronael: We allow also negative premium.  But it should be constrained somehow...

ochaloup:
@kronael can you check the `cargo test` results, please?

settlement-config.yaml
```diff
@@ -1,6 +1,7 @@
 ---
 fee_config:
-  marinade_fee_bps: 800
+  min_fee_bps: 200
+  max_fee_bps: 800
```
ochaloup, line 4: Are we possibly missing the `min_yield_premium_over_ssi_pmpe`?
(probably the issue with the cargo test)

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -298,9 +299,29 @@ pub fn generate_bid_settlements(
...
+                        && total_marinade_active_stake > 0 =>
+                {
+                    let target = ssi + fee_config.min_yield_premium_over_ssi_pmpe;
+                    let staker_yield_pmpe = total_marinade_stakers_rewards
+                        / Decimal::from(total_marinade_active_stake)
+                        * Decimal::ONE_THOUSAND;
+                    let fee_cap = (Decimal::ONE - target / staker_yield_pmpe).max(Decimal::ZERO);
+                    fee_cap.clamp(fee_percentages.min_fee, fee_percentages.max_fee)
+                }
+                _ => fee_percentages.max_fee,
```
ochaloup, line 315: Is it correct that when `total_marinade_active_stake == 0`, we use `max_fee` here?

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -46,24 +49,43 @@ pub struct DaoConfig {
...
+            self.min_fee_bps
+        );
+        ensure!(
+            self.min_fee_bps <= self.max_fee_bps,
+            "min_fee_bps {} exceeds max_fee_bps {} (would push effective fee above configured fee)",
+            self.min_fee_bps,
+            self.max_fee_bps
+        );
+        ensure!(
+            self.min_yield_premium_over_ssi_pmpe.abs() <= dec!(0.1),
```
ochaloup, line 85: This is probably a good thing, as we discussed in another PR discussion. Just thinking we could note it somewhere in a `yaml` comment, maybe?

ochaloup (dismissed):

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: The idea is that it does not really matter what number we use... so we rather use "the default".  Does this check out?

kronael, line 315:
```suggestion
                // does not really matter what we use here, so we use the "default"
                _ => fee_percentages.max_fee,
```

settlement-distributions/bid-distribution/src/settlement_config.rs
kronael: I am not sure, but can add.  My intuition would be do least docs that can then dissynchronize....

kronael:
Added some tests.  And addressed the comments and requests.  Please re-review once more if you can.

settlement-distributions/bid-distribution/src/generators/tests.rs
```diff
@@ -2251,6 +2277,83 @@ fn test_settlement_config_yaml_deserialization() {
...
+        &create_test_settlement_config(),
+        &fee_config,
+        &|pk: &Pubkey| *pk == TEST_PUBKEY_MARINADE,
+        Some(Decimal::try_from(ssi_pmpe).unwrap()),
+    )
+    .unwrap()
+}
+
+#[test]
+fn test_ssi_fee_cap_active_reduces_fee() {
```
ochaloup, line 2345: Possibly to add a test that will check
(a) mixed active+activating stake under the SSI path
and/or (b) activating-only validators with ssi_pmpe = Some(_)

ochaloup (dismissed):

ochaloup (approved):

PR #395, opened 2026-04-21
Add CLI statistics calls and add guide to create an announcement [GEN-7717]

ochaloup:
At commit `bc82913 ("[notifications] replacement for announcements [GEN-7357] (#376)", 2026-04-07)`, the CLI data statistics gathering was mistakenly removed. This PR restores that functionality, introduces a single-purpose API call, removes announcements from the bonds API as they no longer serve a function, and provides guidelines for pushing new announcements.


api/src/rate_limit.rs
```diff
@@ -0,0 +1,143 @@
...
+//! Cloudflare, enforced at the infrastructure layer (security group /
+//! WAF / network ACL). If that invariant ever breaks, a client reaching
+//! the origin directly can spoof `cf-connecting-ip` to a fresh value
+//! per request and obtain a brand-new bucket each time, effectively
+//! bypassing the limiter. This is an **accepted risk**: the code does
+//! not re-enforce the invariant (e.g. by IP-allowlisting Cloudflare
+//! edges) — maintaining an in-code list of CF ranges is operationally
+//! brittle and we rely on the network layer instead.
+//!
+//! # Keyed-store memory
```
kronael, line 20: Might be solvable by hashing the IPs into buckets so that we'd have fixed size and sometimes, more addresses would share a bucket.

api/README.md
```diff
@@ -56,48 +56,8 @@ cargo run --bin api -- --postgres-url "$POSTGRES_URL" \
...
-
-There is also a possibility to add multiple announcements and filter them by operation type. See the available DB columns in [cli announcements sql](../migrations/0005-add-cli-announcements.sql).
-
-To check what is shown in the CLI, run:
-
-NOTE: For CLI installation, [see the doc](../packages/validator-bonds-cli/README.md#npm-installation-details).
 
-```sh
-validator-bonds -u "$RPC_URL" show-bond
+# CLI telemetry write — POST (called by the CLI on every invocation;
```
kronael, line 60: So.. just thinking about this here, if its really not entirely replaceable by the write to Mixpanel...  E.g. we may be able to choose not to track the users in the CLI which would then have the exact same effect as this here.

kronael (approved):

api/src/rate_limit.rs
ochaloup: The best would be to start using `axum` here directly to not do this manual handling. I think I will try to migrate it when I get to a cleaning task.

api/README.md
ochaloup: Yes. I'm not sure how this is handled with the secrets for publishing statistics to Mixpanel, but once we send the statistics directly to Mixpanel, it should be fine, since we won't need to store them in our systems.

PR #405, opened 2026-04-29
[events] emit activating stake fee events

ochaloup:
Enable stake fee events being emitted


ochaloup:
To the commit that enables event emission, I added a fix. Thanks for the review.

kronael (approved):

PR #407, opened 2026-04-30
[events] fix for non defined data [GEN-7600][trivial]

ochaloup:
There are still events that exploit NaN (due to Slonik’s strict behavior). Sanitizing it for JSON compilation.

### DIFF
--- a/packages/bonds-eventing/src/persist-events.ts
+++ b/packages/bonds-eventing/src/persist-events.ts
@@ -3,6 +3,27 @@ import { type DatabasePool, sql, type SerializableValue } from 'slonik'
 import type { BondsEventV1, EmitResult } from './types'
 import type { LoggerWrapper } from '@marinade.finance/ts-common'
 
+// slonik's `sql.jsonb(...)` uses safe-stable-stringify in strict mode, which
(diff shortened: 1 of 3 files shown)


packages/bonds-eventing/src/persist-events.ts
```diff
@@ -3,6 +3,27 @@ import { type DatabasePool, sql, type SerializableValue } from 'slonik'
 import type { BondsEventV1, EmitResult } from './types'
 import type { LoggerWrapper } from '@marinade.finance/ts-common'
 
+// slonik's `sql.jsonb(...)` uses safe-stable-stringify in strict mode, which
+// throws on NaN / ±Infinity. Native `JSON.stringify` (used by the emit step)
+// silently maps them to null — this helper aligns the persisted payload with
+// what the notifications API already receives.
+export function sanitizeForJsonb(value: unknown): unknown {
```
kronael, line 10: Best if we put this to common.  Funny how this works.  Have a pretty much the same function in Python somehwere.  It also has a hook, e.g. the object can define how it parses.  But I don't quite like it as ser-de-like things seem better.

kronael (approved):

PR #412, opened 2026-05-05
[api] skip unknown type variants [trivial]

ochaloup:
When a new type is added and the API is not redeployed, only log an error. Once the API is redeployed, it will load the existing enums from common rs and be able to serve them again.

### DIFF
--- a/api/src/repositories/protected_events.rs
+++ b/api/src/repositories/protected_events.rs
@@ -27,17 +27,48 @@ async fn get_protected_events(
         .await?;
 
     let mut protected_events = vec![];
+    let mut skipped = 0usize;
     while rs.next_row() {
-        protected_events.push(ProtectedEventRecord {
-            epoch: rs.get_i64_by_name("epoch")?.unwrap().try_into()?,
-            amount: rs.get_i64_by_name("amount")?.unwrap().try_into()?,
-            vote_account: Pubkey::from_str(&rs.get_string_by_name("vote_account")?.unwrap())?,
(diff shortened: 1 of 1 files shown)


api/src/repositories/protected_events.rs
```diff
@@ -27,19 +27,50 @@ async fn get_protected_events(
...
+    let mut skipped = 0usize;
     while rs.next_row() {
-        protected_events.push(ProtectedEventRecord {
-            epoch: rs.get_i64_by_name("epoch")?.unwrap().try_into()?,
-            amount: rs.get_i64_by_name("amount")?.unwrap().try_into()?,
-            vote_account: Pubkey::from_str(&rs.get_string_by_name("vote_account")?.unwrap())?,
-            meta: serde_json::from_str(&rs.get_string_by_name("meta")?.unwrap())?,
-            reason: serde_json::from_str(&rs.get_string_by_name("reason")?.unwrap())?,
-        });
+        match parse_row(&rs) {
```
kronael, line 32: Is it possible to just "return it as is" ... e.g. not parse it?  But I guess then the contract would be broken.  I mean.  Is it better to FAIL and break things, or to HIDE the data?

kronael (approved):

api/src/repositories/protected_events.rs
ochaloup: It is possible but I think it just breaks the contract of the enum. The alternative (return raw row) breaks the `ProtectedEventRecord` contract used downstream, and panicking on a single malformed row would block all protected-events refresh until manual intervention. The `log::error!` per row plus the aggregate `log::warn!` already gives loud visibility.

PR #416, opened 2026-05-11
[buildkite] annotate with rpc url path [GEN-7600]

ochaloup:
Pipelines to annotate which RPC_URL they are running with.

### DIFF
--- a/.buildkite/claim-settlements.yml
+++ b/.buildkite/claim-settlements.yml
@@ -47,7 +47,9 @@ steps:
         slack_feed=$(./scripts/bonds-slack-feed.sh "$$claim_type")
         buildkite-agent meta-data set --redacted-vars='' slack_feed "$$slack_feed"
       - |
+        printable_rpc_url=$(./scripts/printable-rpc-url.sh "$$RPC_URL")
         echo "Claim Type: '$$claim_type'/'$$config_pubkey', Slack feed: '$$slack_feed'"
+        echo "# RPC_URL='$$printable_rpc_url'"
 
   - label: ":memo: Annotate retry parameters"
     depends_on: setup-env-claim
@@ -55,6 +57,7 @@ steps:
     commands:
       - claim_type=${CLAIM_TYPE:-$(buildkite-agent meta-data get claim_type)}
       - slack_feed=$(buildkite-agent meta-data get slack_feed)
+      - printable_rpc_url=$(./scripts/printable-rpc-url.sh "$$RPC_URL")
       - |
         cat <<- EOF | buildkite-agent annotate --priority 6 --style info --context "general-info"
         ### Retry environment variables
@@ -64,6 +67,7 @@ steps:
         CLAIM_TYPE=$$claim_type
         SLACK_FEED=$$slack_feed
         NOTIFY_FEED=${NOTIFY_FEED:-false}
+        # RPC_URL='$$printable_rpc_url'
         \`\`\`
         EOF
 
--- a/.buildkite/close-settlements.yml
+++ b/.buildkite/close-settlements.yml
@@ -47,7 +47,9 @@ steps:
       slack_feed=$(./scripts/bonds-slack-feed.sh "$$claim_type")
       buildkite-agent meta-data set --redacted-vars='' slack_feed "$$slack_feed"
     - |
+      printable_rpc_url=$(./scripts/printable-rpc-url.sh "$$RPC_URL")
       echo "Claim Type: '$$claim_type'/'$$config_pubkey'"
+      echo "# RPC_URL='$$printable_rpc_url'"
 
   - label: ":memo: Annotate retry parameters"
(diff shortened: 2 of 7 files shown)


shejby (changes requested):
Concerns

  - Userinfo leak. The regex ^([a-zA-Z][a-zA-Z0-9+.-]*://[^/]+).* captures everything between :// and the first /, which includes user:pass@ if present. An RPC URL of the form
  https://user:secret@host/path would render as https://user:secret@host/... — secret preserved verbatim. Solana RPC providers don't typically use Basic-auth userinfo, but it's a
  single regex away from being safe regardless:
  sed -E 's#^([a-zA-Z][a-zA-Z0-9+.-]*://)([^@/]+@)?([^/]+).*#\1\3#'
  - Worth adding even as defense-in-depth, since the cost is zero and "we never use userinfo" is an invariant nobody is enforcing.
  - *waypoint* substring match is broader than intended. Any hostname containing waypoint (incl. waypoint-impersonator.attacker.com) is treated as trusted-internal and returned
  verbatim. The RPC URL is ops-controlled, not user-controlled, so the practical exposure requires already-compromised config — but a tighter match (anchored suffix like
  *.waypoint.marinade.finance, or matching waypoint. as a host-prefix rather than substring) would close the gap without losing flexibility. If the intent is to catch aliases like
  internal-waypoint or waypoint-stage, an enumerated allowlist would be cleaner.
  - Edge case: no path. https://api.example.com (no path) matches the regex with .* capturing empty string, so the output is https://api.example.com/... — adds a misleading /...
  suffix when there was no path. Not a security issue (no secrets exist), just visually misleading. Could be addressed with a conditional: only append /... if the original URL had a
  path or query.
  - No set -euo pipefail. The script tolerates a sed parse failure silently — the output would be the original input, then prefixed/handled by the rest of the logic. Failure mode is
  "garbled annotation," not "leaked secret," so not blocking, but a one-line addition.

  Smaller Observations

  - #!/bin/bash rather than #!/usr/bin/env bash. Buildkite agents have bash at /bin/bash, so fine; env form would be more portable.
  - No test plan / no tests. For a sanitizer that handles secret-carrying URLs, even a small inline bash test (or a couple of bats cases) would be cheap regression insurance. Concrete
   cases worth pinning:
    - https://api.helius.xyz/v0/SECRET-KEY/abc → https://api.helius.xyz/...
    - https://api.helius.xyz?api-key=SECRET → https://api.helius.xyz/...
    - https://user:pw@host/path → https://host/... (after userinfo fix)
    - https://my-waypoint.internal/v1/x?strategy=race → returned as-is
    - empty / unset → (unset)
  - $$RPC_URL correctly uses Buildkite's double-$ escape for runtime shell interpolation. Consistent with the rest of these pipelines.
  - Annotation form # RPC_URL='...' renders as a shell-comment line inside a fenced code block. Reads naturally.

  Verdict

  Useful observability addition; threat model is right. Two soft asks before merge — both one-liner script changes:

  1. Strip userinfo in the sanitizer regex (defense-in-depth against Basic-auth-style URLs).
  2. Tighten the *waypoint* match or document the intent (substring-anywhere vs anchored-suffix).

ochaloup:
@shejby thanks for the review. Fixed or/and explained below. Can you please re-review?

- Userinfo leak in regex (user:pass@host preserved verbatim)
  Fixed. Regex now extracts scheme + host only, dropping the optional
  userinfo@ group. Verified: https://user:pass@host:8899/path → https://host:8899/...
- waypoint substring match too broad (matches *waypoint*, incl. impersonators)
  Declined. Minor risk in practice — RPC URL is ops-controlled, not user-controlled.
- No-path URLs get a misleading "/..." suffix (https://api.example.com → .../...)
  Fixed. Now only appends "/..." when the original URL has a path/query/fragment.
  Verified: https://api.example.com → https://api.example.com (no suffix);
  https://api.example.com/ → https://api.example.com/... (suffix kept).
- No `set -euo pipefail`
  Declined. The earlier silent-sed-failure path is gone (replaced by bash regex),
  so the concrete failure mode no longer applies here.
- #!/bin/bash vs #!/usr/bin/env bash
  Fixed. Switched to /usr/bin/env bash for portability.
- No tests for the sanitizer
  Declined. No existing test harness for shell scripts in this repo; out of scope.
- $$RPC_URL Buildkite escape / annotation form
  No action — observations only, already correct.

shejby (changes requested):
I believe this PR is not [GEN-7600][trivial].
Finalize Waypoint / trivial

ochaloup:
@shejby Removing `trivial` as it seems misleading. Considering this as part of the waypoint, as I would like to have a better idea of what RPC config was used in the pipeline. Recently, we were switching `RPC_URL` in the Buildkite environment back and forth, so having a clear understanding of the setup for a particular run would help.

shejby (approved):

PR #418, opened 2026-05-13
Switch to apy-api SSI [GEN-7676]

kronael:
Switch SSR source to apy.marinade.finance.


kronael:
If we merge this now, it should just not fire until the epoch is available.  ... or rather it will be failing repeatedly.  What would be the ideal protocol?  IMHO we should wait for the epoch data to be filled, but then we'd need to FAIL if its like slot 250000 and still no data.

settlement-distributions/bid-distribution/src/bin/cli.rs
```diff
@@ -60,6 +61,10 @@ struct Args {
     /// Output path for protected events collection JSON (PSR only)
     #[arg(long, env)]
     output_protected_event_collection: Option<String>,
+
+    /// Base URL of the apy-api service (used to fetch SSR pmpe for the scoring epoch)
```
ochaloup: Can we maybe change the comment to `... fetch SSI/SSR pmpe ...` so it's clear that SSI and SSR are the same?

Overall, in the whole PR, it would be good to use both terms in comments and the single term for variables to be a bit more consistent.

settlement-distributions/bid-distribution/src/apy_api.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[derive(Deserialize)]
+struct EpochPmpeResponse {
+    epochs: Vec<EpochPmpeEntry>,
+}
+
+pub fn fetch_ssr_pmpe(apy_api_url: &str, epoch: u64) -> anyhow::Result<Decimal> {
+    let url = format!("{apy_api_url}/v1/epoch-pmpe/ssr");
+    info!("Fetching SSR pmpe for epoch {epoch} from {url}");
+    let resp: EpochPmpeResponse = reqwest::blocking::get(&url)?.error_for_status()?.json()?;
```
ochaloup: Maybe add `.timeout()` so it doesn't block forever?
And maybe to add `.with_context` to get more info on failure?

settlement-distributions/bid-distribution/src/generators/tests.rs
ochaloup: nitpick: `Optional` was removed so `Some` comment is stale

ochaloup (changes requested):
I like the processing, and I like the fact that it comes from the API.

* My main concern is that we need to update the scheduler so this is handled correctly. The scheduler decides when to run the pipeline calculation — `scheduler-bidding.yml` — and it should only run once all required data is available.

    I asked Claude to sketch what the code could look like — please verify if it matches the other verifications in the scheduler:


```
  - |
    set -o pipefail
    # ... existing ssr-unrelated lines ...
    ssr_json=$(curl -fsSL --max-time 15 "https://apy.marinade.finance/v1/epoch-pmpe/ssr") || exit 23
    max_available_ssr_epoch=$(echo "$$ssr_json" | jq '[.epochs[].epoch] | max // 0')
    echo "Bidding:: > max_available_ssr_epoch: $$max_available_ssr_epoch"
  - |
    # ... existing meta-data set lines ...
    buildkite-agent meta-data set max_available_ssr_epoch "$$max_available_ssr_epoch"

  In the "Schedule" block (line 43), extend the gate:

  - max_available_ssr_epoch=$(buildkite-agent meta-data get max_available_ssr_epoch)
  - |
    if (( $$max_available_etl_epoch >= $$max_available_epoch )) \
       && (( $$max_available_ssr_epoch >= $$max_available_epoch )); then
      ...
```
* Another thing is that I was trying to explicitly define the used APIs in the pipelines to make it easier to understand what is needed for running. I believe it would be good to add `env: APY_API_URL=...` at the start of `prepare-bid-distribution.yml`.
  Similarly, maybe the README should be updated as well (not sure, to be honest), since there is another API used for pipeline calculation.

* As some follow-up PR (not here), it would be good to consider adjusting the testing script `scripts/regression-test-settlements.sh` around the APY API processing.

settlement-distributions/bid-distribution/src/apy_api.rs
kronael: There should be some TCP or HTTP timeout no?

kronael: I see there is none, ok.  Added that.

.buildkite/scheduler-bidding.yml
```diff
@@ -26,11 +26,14 @@ steps:
 
         max_available_epoch=$(gcloud storage ls "${gs_bucket_snapshots}/**/stakes.json" | awk -F / '{print $$4}' | sort -nr | head -n 1 || exit 2) || exit 2
         max_available_etl_epoch=$(gcloud storage ls "$gs_bucket_etl" | awk -F / '{print $$4}' | sort -nr | head -n 1) || exit 22
-        echo "Bidding:: > max_processed_epoch_bid: $$max_processed_epoch_bid, max_available_epoch: $$max_available_epoch, max_available_etl_epoch: $$max_available_etl_epoch"
+        ssr_json=$(curl -fsSL --max-time 15 "https://apy.marinade.finance/v1/epoch-pmpe/ssr") || exit 23
```
ochaloup: Would it be possible to place the URL into the `env:` section, please?

settlement-distributions/bid-distribution/src/apy_api.rs
```diff
@@ -0,0 +1,32 @@
...
+struct EpochPmpeResponse {
+    epochs: Vec<EpochPmpeEntry>,
+}
+
+// SSI and SSR refer to the same network-wide Solana Staking Index/Rate (pmpe).
+pub fn fetch_ssr_pmpe(apy_api_url: &str, epoch: u64) -> anyhow::Result<Decimal> {
+    let url = format!("{apy_api_url}/v1/epoch-pmpe/ssr");
+    info!("Fetching SSI/SSR pmpe for epoch {epoch} from {url}");
+    let client = reqwest::blocking::Client::builder()
+        .timeout(HTTP_TIMEOUT)
```
ochaloup, line 25: nitpick idea: adding `anyhow::Context` and chain `.with_context(|| format!("apy-api request failed for {url}"))?` on both `.send()` and `.json()`

settlement-distributions/bid-distribution/src/apy_api.rs
```diff
@@ -0,0 +1,32 @@
...
+}
+
+#[derive(Deserialize)]
+struct EpochPmpeResponse {
+    epochs: Vec<EpochPmpeEntry>,
+}
+
+// SSI and SSR refer to the same network-wide Solana Staking Index/Rate (pmpe).
+pub fn fetch_ssr_pmpe(apy_api_url: &str, epoch: u64) -> anyhow::Result<Decimal> {
+    let url = format!("{apy_api_url}/v1/epoch-pmpe/ssr");
```
ochaloup: Thinking about this joining `/` that may cause having `//`. What to use something like `reqwest::Url::join`?

ochaloup (dismissed):
It all seems fine to me. For code clarity, I would prefer to see the URL defined as `env`, but it can also be fixed in some follow-up cleanup PR.

settlement-distributions/bid-distribution/src/apy_api.rs
kronael: That is ofc a good idea! ... (adding it to the skill)

settlement-distributions/bid-distribution/src/apy_api.rs
kronael: Claude was telling me this too.  How could that happen, who would setup the base URL with a trailing slash?  But ok, adding it somehow so that its defensive.

settlement-distributions/bid-distribution/README.md
```diff
@@ -13,9 +13,14 @@ pipeline_dir="/tmp/bond-pipeline-${epoch}" && mkdir -p "$pipeline_dir"
 # generated by @marinade-finance/validator-bonds/snapshot-parser-cli
 gcloud storage cp "gs://$bucket/$epoch/stakes.json" "${pipeline_dir}/stakes.json"
 
-# Input data comes from ds-sam-pipeline
+# Input data comes from ds-sam-pipeline (no longer carries ssr_pmpe; SSI/SSR is fetched from apy-api).
```
kronael:
```suggestion
# Input data comes from ds-sam-pipeline.
```

settlement-distributions/bid-distribution/src/apy_api.rs
```diff
@@ -0,0 +1,39 @@
...
+    let client = reqwest::blocking::Client::builder()
+        .timeout(HTTP_TIMEOUT)
+        .build()?;
+    let resp: EpochPmpeResponse = client
+        .get(url.as_str())
+        .send()
+        .with_context(|| format!("apy-api request failed for {url}"))?
+        .error_for_status()?
+        .json()
+        .with_context(|| format!("apy-api request failed for {url}"))?;
```
ochaloup: nitpick: the `json()` should probably have a different context than the fetch call

settlement-distributions/bid-distribution/src/apy_api.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+#[derive(Deserialize)]
+struct EpochPmpeResponse {
+    epochs: Vec<EpochPmpeEntry>,
+}
+
+// SSI and SSR refer to the same network-wide Solana Staking Index/Rate (pmpe).
+pub fn fetch_ssr_pmpe(apy_api_url: &str, epoch: u64) -> anyhow::Result<Decimal> {
+    let url = reqwest::Url::parse(apy_api_url)?.join("/v1/epoch-pmpe/ssr")?;
```
ochaloup: I was checking the `join`, and it seems to be more complicated. I was thinking it worked somehow more cleverly than it seems it does.

```
Full truth table for <base>.join("a/b/c"):

https://x.com               → https://x.com/a/b/c
https://x.com/              → https://x.com/a/b/c
https://x.com/something     → https://x.com/a/b/c        (drops "something")
https://x.com/something/    → https://x.com/something/a/b/c
```

Claude pointed down another option[1], but that's kind of lengthy.

[1]
```
let mut url = reqwest::Url::parse(apy_api_url)?;
url.path_segments_mut()
  .map_err(|_| anyhow::anyhow!("invalid apy-api base URL: {apy_api_url}"))?
  .pop_if_empty()
  .extend(&["v1", "epoch-pmpe", "ssr"]);
```

ochaloup (dismissed):

settlement-distributions/bid-distribution/src/apy_api.rs
kronael: Really?  Seems like a really complicated thing to ship a string concatenation :D

kronael: I did reverrt it to the original Claude thing.  Which I did not like, but its ... the best of these :100:

settlement-distributions/bid-distribution/src/apy_api.rs
```diff
@@ -0,0 +1,39 @@
...
+    let client = reqwest::blocking::Client::builder()
+        .timeout(HTTP_TIMEOUT)
+        .build()?;
+    let resp: EpochPmpeResponse = client
+        .get(&url)
+        .send()
+        .with_context(|| format!("apy-api request failed for {url}"))?
+        .error_for_status()?
+        .json()
+        .with_context(|| format!("apy-api request failed for {url}"))?;
```
kronael:
```suggestion
        .with_context(|| format!("apy-api json parsing failed for {url}"))?;
```

ochaloup (approved):

PR #421, opened 2026-05-18
Set min_fee=0 [GEN-7676]

kronael:
Mostly because as its per validator it seems we are not giving back enough as this might not be ideal, but will study this later.  For now we need to hotfix this.

### DIFF
--- a/settlement-config.yaml
+++ b/settlement-config.yaml
@@ -1,6 +1,6 @@
 ---
 fee_config:
-  min_fee_bps: 200
+  min_fee_bps: 0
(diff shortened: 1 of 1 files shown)


settlement-config.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 fee_config:
-  min_fee_bps: 200
+  min_fee_bps: 0
   max_fee_bps: 800
```
kronael:
```suggestion
  max_fee_bps: 700
```

janlegner (approved):

PR #423, opened 2026-05-18
CLI mixpanel eventing [GEN-7717]

ochaloup:
Analytics going to a company-wide analytics solution instead of an internal DB to make the data easier to understand.


ochaloup:
Changing the telemetry from CLI to be sent to Mixpanel.

packages/validator-bonds-cli-core/__tests__/cliUsage-injection.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+  mkdtempSync,
+  readFileSync,
+  rmSync,
+} from 'node:fs'
+import { tmpdir } from 'node:os'
+import { join, resolve } from 'node:path'
+
+const PACKAGE_DIR = resolve(__dirname, '..')
+const REPO_ROOT = resolve(PACKAGE_DIR, '..', '..')
+const INJECT_SCRIPT = join(REPO_ROOT, 'scripts', 'inject-mixpanel-token.js')
```
kronael, line 14: `resolve` is unnecessary here — `__dirname` is already absolute, so plain `join` suffices. Drops `resolve` from the `'node:path'` import too:

```ts
import { join } from 'node:path'

const PACKAGE_DIR   = join(__dirname, '..')
const REPO_ROOT     = join(__dirname, '..', '..', '..')
const INJECT_SCRIPT = join(REPO_ROOT, 'scripts', 'inject-mixpanel-token.js')
const DIST_CLIUSAGE = join(PACKAGE_DIR, 'dist', 'src', 'cliUsage.js')
```

Optional, more semantic alternative — anchor on `package.json` so the spec file can move within `__tests__/` without breaking the path math:

```ts
const PACKAGE_DIR = dirname(require.resolve('../package.json'))
const REPO_ROOT   = dirname(dirname(PACKAGE_DIR))
```

Either is fine; just `resolve` → `join` is the lowest-friction change.

packages/validator-bonds-cli-core/__tests__/cliUsage-injection.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+    const tmpFile = join(tmpDistSrc, 'cliUsage.js')
+    copyFileSync(DIST_CLIUSAGE, tmpFile)
+
+    execFileSync('node', [INJECT_SCRIPT], {
+      cwd: tmpDir,
+      env: { ...process.env, MIXPANEL_TOKEN: 'spec-injected-token' },
+    })
+
+    const after = readFileSync(tmpFile, 'utf-8')
+    expect(after).toContain("BUILD_TIME_MIXPANEL_TOKEN = 'spec-injected-token'")
```
kronael: claude: brittle assertion — ties the test to exact TypeScript compiler output formatting (quotes, spacing). Use `after.includes('spec-injected-token')` and `!after.includes('__MIXPANEL_TOKEN_PLACEHOLDER__')` to be format-agnostic.

packages/validator-bonds-cli/__tests__/test-validator/cli-mixpanel.spec.ts
```diff
@@ -0,0 +1,341 @@
...
+    ]).toHaveMatchingSpawnOutput({
+      code: 0,
+      signal: '',
+      stdout: new RegExp(configPubkey.toBase58()),
+    })
+
+    expect(capturedEvents).toHaveLength(0)
+  })
+
+  it('emits zero events when no Mixpanel token is configured', async () => {
```
kronael, line 267: claude: environment-dependent test — `MIXPANEL_TOKEN_TEST: ''` falls back to `BUILD_TIME_MIXPANEL_TOKEN`. Against an injected binary that's a real token, so events fire and `toHaveLength(0)` fails. Use `MIXPANEL_TOKEN_TEST: '__MIXPANEL_TOKEN_PLACEHOLDER__'` explicitly to pin the "disabled" sentinel.

scripts/inject-mixpanel-token.js
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const token = process.env.MIXPANEL_TOKEN
+if (!token || token.length === 0) {
+  die('MIXPANEL_TOKEN env var must be set and non-empty for publish builds')
+}
+if (token === PLACEHOLDER) {
+  die('MIXPANEL_TOKEN must not equal the placeholder value')
+}
+if (!/^[A-Za-z0-9._-]+$/.test(token)) {
```
kronael: claude: overly permissive token regex — allows dots and hyphens, but Mixpanel tokens are hex strings. If the format is stable, tighten to `^[A-Za-z0-9]+$` to catch misconfigurations earlier.

kronael:
claude: \`setAccountField\` missing on 8 commands — all fall back to the generic \`account\` field in Mixpanel instead of a typed field name.

Files not touched in this PR that need the same treatment as \`closeSettlement\`, \`configureConfig\`, etc.:

| Command | File | Suggested field |
|---|---|---|
| \`show-bond\` | \`src/commands/show.ts:109\` | \`vote_account\` |
| \`configure-bond\` | \`src/commands/manage/configureBond.ts:43\` | \`vote_account\` |
| \`fund-bond\` | \`src/commands/manage/fundBond.ts:36\` | \`vote_account\` |
| \`fund-bond-sol\` | \`src/commands/manage/fundBondWithSol.ts:42\` | \`vote_account\` |
| \`mint-bond\` | \`src/commands/manage/mintBond.ts:25\` | \`vote_account\` |
| \`init-withdraw-request\` | \`src/commands/manage/initWithdrawRequest.ts:44\` | \`vote_account\` |
| \`cancel-withdraw-request\` | \`src/commands/manage/cancelWithdrawRequest.ts:30\` | \`vote_account\` |
| \`claim-withdraw-request\` | \`src/commands/manage/claimWithdrawRequest.ts:35\` | \`vote_account\` |

\`bond_account\` is not in the \`AccountField\` union — add it or use \`vote_account\` as the primary identity.

---

claude: stale comment in \`migrations/0005-add-cli-announcements.sql\` — still reads "still in use — written by \`/v1/cli-usage\`". The endpoint is now deleted. Add a \`COMMENT ON TABLE cli_usage IS 'DEPRECATED: no longer written to, telemetry moved to Mixpanel'\` in a new migration.

kronael (changes requested):

kronael (dismissed):
LGTM.  Claude found something, so flagging, but have not much of an opinion on that.

ochaloup:
@kronael thanks for the points. I'm reviewing them.

The Claude point

```text
`setAccountField` missing on 8 commands — all fall back to the generic `account` field in Mixpanel instead of a typed field name.
```

was something I previously disagreed with, but having all this prepared and after revisiting the Mixpanel docs, I believe it is a valid point and can make the processing more generic (and nicer).

I will do a bigger refactoring and ask you for a re-review afterward.

scripts/inject-mixpanel-token.js
ochaloup: considered but not a big issue, rather skipping

packages/validator-bonds-cli/__tests__/test-validator/cli-mixpanel.spec.ts
ochaloup: that does not seem a big issue

packages/validator-bonds-cli-core/__tests__/cliUsage-injection.spec.ts
ochaloup: good point but a test issue

ochaloup:
@kronael can you re-review please :pray:

migrations/0008-deprecate-cli-usage-table.sql
kronael, line 1: Is there a way to feed the data to mixpanel?  Easily?  Or like export it and give it to maks - he might just safe-keep it then.

packages/validator-bonds-cli-core/src/commands/mainCommand.ts
```diff
@@ -105,14 +115,30 @@ export function launchCliProgram({
...
+      new Option('--mix-proxy-url <url>', 'Override Mixpanel proxy URL')
+        .env('MIX_PROXY_URL')
+        .default(DEFAULT_MIX_PROXY_URL)
         .hideHelp(),
     )
 
   installAdditionalOptions(program)
 
+  let pendingCompletion:
+    | {
```
kronael: nit: I'd create a named type for this... but its.. just a nit.

packages/validator-bonds-cli-core/src/cliUsage.ts
```diff
@@ -1,6 +1,54 @@
...
+
+import type { PublicKey } from '@solana/web3.js'
+import type { Command } from 'commander'
 import type { Logger } from 'pino'
 
-const CLI_USAGE_TIMEOUT_MS = 1500
+const MIXPANEL_TIMEOUT_MS = 1500
+
+// Array#join avoids constant-folding, so the injector rewrites only BUILD_TIME_MIXPANEL_TOKEN.
+const MIXPANEL_TOKEN_PLACEHOLDER = [
```
kronael, line 15: :D ... wow

kronael (dismissed):

packages/validator-bonds-cli-core/src/cliUsage.ts
kronael: I mean.  Very ingenious.

ochaloup: That's the Claude... unfortunately

migrations/0008-deprecate-cli-usage-table.sql
ochaloup: Yes, but I'm not sure how easily. The current data is not capturing the amount of information the new setup should generate.
As we currently don't know what we want to read from the data, I will leave the potential migration for later if it turns out the available data has some value.

packages/validator-bonds-cli-core/src/commands/mainCommand.ts
ochaloup: I really like this point. This is too vague. I'm adding the type for it.

ochaloup:
@kronael I did a small update based on your review point. Can you please re-review? :pray:

kronael (dismissed):

ochaloup:
@kronael many thanks for all the comments on this PR. After merging the fixes from https://github.com/marinade-finance/validator-bonds/pull/430, I needed to resolve conflicts and update the CLI version so both could be published. I rebased and force-pushed the branch. Would you be so kind as to re-review the updated state?

ochaloup:
@kronael I fixed the issues from Copilot and added a few more nitpick fixes from the previous review. Could you please review the code now?

packages/validator-bonds-cli-core/src/cliUsage.ts
kronael: With these things... its the kind of street-smarts I love about him :D

kronael (dismissed):

ochaloup:
@kronael can I ask for a re-review :pray: ? I updated the PR to get lint passed + fix from copilot review

pnpm-workspace.yaml
```diff
@@ -7,3 +7,11 @@ allowBuilds:
   node-hid: true
   unrs-resolver: true
   utf-8-validate: true
+overrides:
```
kronael, line 10: Would just check if there are not some Claude inventions and are really needed.  (I have no clue)

kronael (approved):

pnpm-workspace.yaml
ochaloup: This is something I fought with, and I'm still not fully clear on how it works across versions. But this is caused by the pnpm version switch. pnpm 11 ignores all related settings from `package.json`, and everything needs to be configured in the workspace YAML. pnpm 10 deprecated that handling but was still capable of reading `package.json`. pnpm 11 simply removed that handling altogether.

PR #424, opened 2026-05-18
[pipelines] adjust verify processing [GEN-7963]

ochaloup:
The verify pipeline started failing too often and creating noise that is hard to determine whether it is wrong or not. Making it less verbose while still keeping it relevant to inform about issues.

### DIFF
--- a/scripts/parse-verify-alerts.sh
+++ b/scripts/parse-verify-alerts.sh
@@ -1,9 +1,12 @@
 #!/bin/bash
 set -euo pipefail
 
-# Parse settlement verification alerts from JSON report of 'verify-settlement' command
-# Usage: parse-verify-alerts.sh -f <verify-report.json> [-n <non-funded-items-to-report>] [-e <non-existing-items-to-report>] [-m <max-display-items>]
-
+# Parse settlement verification alerts from JSON report of 'verify-settlement' command.
+# Non-funded settlements are classified per epoch; any epoch crossing a threshold
+# raises an alert that is posted to Slack and fails the build:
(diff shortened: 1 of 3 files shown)


scripts/parse-verify-alerts.sh
```diff
@@ -105,59 +127,97 @@ if [ "$non_verified_count" -gt 0 ]; then
...
+        non_funded_epochs_total=$((non_funded_epochs_total + 1))
+
+        sol=$(awk -v l="$lamports" 'BEGIN { printf "%.4f", l / 1e9 }')
+
+        if [ "$count" -ge "$NON_FUNDED_PER_EPOCH" ] || [ "$lamports" -ge "$sol_threshold_lamports" ]; then
+            non_funded_epochs_alert=$((non_funded_epochs_alert + 1))
+            non_funded_settlements_alerting=$((non_funded_settlements_alerting + count))
+            is_to_alert=true
+            echo "    epoch=$epoch count=$count sol=$sol ALERT" >&2
+            non_funded_summary="${non_funded_summary}Epoch ${epoch}: ${count} settlements / ${sol} SOL
```
martinkrecek: (non-blocking nitpick)
If I'm reading this line right, it appends a newline because of the closing `"` sign being placed on the next line. No issue with keeping this, I'd just consider using an explicit `$'\n'` on the same line after the closing `"` to append a newline to avoid making it essentially depend on the formatting.

martinkrecek (dismissed):

ochaloup:
@martinkrecek I updated based on your review. Can you please re-review :pray:

martinkrecek (approved):

PR #428, opened 2026-05-21
20260521 simulation [GEN-7676]

kronael:
Adds scripts/fee-sensitivity.ts, a bun TypeScript script that simulates
  bid-distribution-cli across a range of epochs at multiple fee tiers.

  For each (epoch, fee) pair it patches settlement-config.yaml into a temp file,
  runs the CLI, and reports post-fee pmpe and APY — both as actually charged
  (adj) and at the full fee rate (max). Fetches SSR epoch timing for APY conversion.


kronael:
Sorry for the tests.  Fixing this now.  Should be something rather trivial.

ochaloup (dismissed):
Just to match the linting rules

ochaloup (approved):

PR #430, opened 2026-05-21
[cli] better error handling on wrong rpc url + defaulting to show funding [GEN-8002]

ochaloup:


packages/validator-bonds-cli-core/src/errorTranslators.ts
kronael, line 1: Is this just taken from somewhere or new?  If new... possibly a common-grade thing?

packages/validator-bonds-cli-core/src/errorTranslators.ts
```diff
@@ -0,0 +1,217 @@
...
+  const seen = new Set<unknown>()
+  while (current && !seen.has(current)) {
+    seen.add(current)
+    chain.push(current)
+    current = (current as { cause?: unknown }).cause
+  }
+  return chain
+}
+
+function findInCauseChain(
```
kronael: Seems like this should just be used in-line.  As it increases the LOC but does not provide any useful abstraction.

packages/validator-bonds-cli-core/src/errorTranslators.ts
```diff
@@ -0,0 +1,217 @@
...
+  rpcEndpoint?: string
+  txArgs?: ExecuteTxParams
+}
+
+type Translator = (
+  err: unknown,
+  ctx: TranslateCtx,
+) => CliCommandError | undefined
+
+export function buildRpcRemediationMsg(prefix: string): string {
```
kronael, line 19: We provide some RPC for the web.  Could we make this a default also in the CLI so that users will have a simpler world?

packages/validator-bonds-cli-core/src/errorTranslators.ts
```diff
@@ -0,0 +1,217 @@
...
+    valueName: 'fee-payer/rent-payer',
+    value: `${feePayerInfo}, balance: ${insufficientMatch[1]} lamports, needed: ${insufficientMatch[2]} lamports`,
+    msg:
+      'The fee payer or rent payer account does not have enough SOL to cover the transaction fees and rent.' +
+      ' Make sure the account is funded before executing the transaction.',
+    cause: err,
+  })
+}
+
+const KNOWN_ERROR_TRANSLATORS: Translator[] = [
```
kronael, line 197: Could this be made into a decorator, where decorating with "@error" (or better) would register it with the list?

packages/validator-bonds-cli-institutional/README.md
```diff
@@ -52,9 +52,9 @@ validator-bonds-institutional init-bond --vote-account <vote-account-address> \
 # number of SOL is transferred to a stake account that is assigned under Validator Bonds program
 validator-bonds-institutional fund-bond-sol <vote-account-address> --from <wallet-keypair> --amount <1 SOL for every 1,000 SOL staked>
 
-# STEP 3: SHOW BOND DATA
+# STEP 3: SHOW BOND DATA (funding info is included automatically for a specific account)
```
kronael: Possibly just a "Claude"-reference-to-history which is not necessary?

packages/validator-bonds-cli-institutional/README.md
```diff
@@ -123,9 +123,11 @@ validator-bonds-institutional configure-bond --authority <recipient-wallet-keypa
 
 ```sh
 RPC_URL=<url-to-solana-rpc-node>
-validator-bonds-institutional -u$RPC_URL show-bond <bond-or-vote-account-address> --with-funding --verbose
+validator-bonds-institutional -u$RPC_URL show-bond <bond-or-vote-account-address> --verbose
 ```
 
+Funding information is included automatically when querying a specific account; for filter-based listings, opt in with `--with-funding`.
```
kronael, line 129: This one seems enough.

kronael (dismissed):
Mostly nits.  Its good work.

packages/validator-bonds-cli-core/src/index.ts
```diff
@@ -4,5 +4,6 @@ export * from './cliUsage'
 export * from './commands'
 export * from './computeUnits'
 export * from './context'
+export * from './errorTranslators'
```
kronael: 🤖 claude: `translateRpcConnectivityError` and `translateRpcRateLimitError` are marked "Not part of cli-core public API; treat as internal" in their JSDoc, but this `export *` puts them on the public package surface. Once exported, callers can depend on them. Either drop the internal comment, or list named exports so only `translateKnownError`, `buildRpcRemediationMsg`, and `resolveRpcEndpoint` are public.

packages/validator-bonds-cli-core/src/errorTranslators.ts
```diff
@@ -0,0 +1,217 @@
...
+
+  return undefined
+}
+
+// Exported solely for unit tests. Treat as internal.
+export const translateRpcRateLimitError: Translator = (err, ctx) => {
+  const endpoint = ctx.rpcEndpoint ?? 'the configured RPC endpoint'
+  const code = getRpcErrorCode(err)
+  const message = err instanceof Error ? err.message : ''
+  if (code === -32005 || code === -32429 || /429/.test(message)) {
```
kronael: 🤖 claude: `/429/` matches any substring containing the digits 4-2-9 (e.g. `"balance: 14290"`, `"request #1429"`). The other checks use exact numeric codes; this one is inconsistent. Consider `/\b429\b/` or `/status[: ]+429/i` to restrict it to HTTP 429 status strings.

packages/validator-bonds-cli-core/src/commands/show.ts
```diff
@@ -338,53 +336,35 @@ export async function showBond({
...
     }
 
-    if (withFunding) {
-      const configAccount = config ?? bondData.account.data.config
-      const bondFunding = await getBondsFunding({
-        program,
-        configAccount,
-        bondAccounts: [address],
-        voteAccounts: [bondData.account.data.voteAccount],
+    // single-account query always includes funding info
```
kronael: 🤖 claude: `--with-funding` is now silently ignored for single-account queries — the fetch is unconditional from here. A user who explicitly omits the flag to avoid rate-limited RPC calls will still hit them with no indication why. A `logger.debug` line noting that funding is always fetched for single-account mode would make this observable.

packages/validator-bonds-cli-core/src/commands/mainCommand.ts
```diff
@@ -195,16 +196,30 @@ export function launchCliProgram({
...
     (err: Error) => {
+      const originalErr = err
+      let rpcEndpoint: string | undefined
+      try {
+        rpcEndpoint = getCliContext().provider.connection.rpcEndpoint
+      } catch (_e) {
+        // context not yet set (error happened before preAction completed)
+      }
+      try {
+        translateKnownError(err, { rpcEndpoint })
```
kronael: 🤖 claude: `translateKnownError` is typed `never` (always throws), so `try { translateKnownError(...) } catch (translated) { err = translated }` uses try/catch as a return-value channel — the catch is the only real exit path. Works correctly, but surprises readers who expect the try body to sometimes succeed. A thin wrapper that returns instead of throwing would make the intent clearer.

packages/validator-bonds-sdk/src/web3.js/voteAccount.ts
```diff
@@ -65,7 +65,7 @@ export async function findVoteAccountByIdentity({
   ) {
     logDebug(
       logger,
-      `Found ${accounts.length} (${accounts.map(a => a.pubkey.toBase58()).join(', ')}) vote accounts for identity ${identity.toBase58()}.` +
+      `Found ${accounts.length} (${accounts.map(a => a.pubkey.toBase58()).join(', ')}) vote accounts for identity ${identity.toBase58()}. ` +
```
kronael, line 68: 🤖 claude: Unrelated whitespace-only change (trailing space added before the next string literal). No behavior impact; just adds noise to the diff.

packages/validator-bonds-cli-core/src/errorTranslators.ts
ochaloup: That's a good point and I will search what can be don in this matter.

packages/validator-bonds-cli-core/src/errorTranslators.ts
ochaloup: This is not a good fit for common in its current shape. It works only with web3.js 1.98.x and Anchor processing, a bit specifically for the bonds CLI. We would need a bit more rework here. I do not see a strong reason to make this generic so far, so I will leave it for some future refactoring.

packages/validator-bonds-cli-core/src/errorTranslators.ts
ochaloup: Maybe, though I’m not sure if it is a good fit here. I’m leaving it for future refactoring, which in the decorator case would have to be pretty substantial.

ochaloup:
@kronael thanks for the comments. I updated the PR a little bit and replied to your points. I like the rationale behind all of that, but for some things (decorators) I'm not convinced they are a good (or possible) fit for the current code, and others need a bit more work (common error handler), so both categories are left out of scope for this PR.
Can I ask for re-review :pray: 

(and I rebased and forced-pushed )

kronael (dismissed):

kronael:
The Claude coments (should;ve been there from before) ... Please ignore if not immediately apparently relevant.

packages/validator-bonds-cli-core/src/errorTranslators.ts
kronael: Oh interesting.  I was thinking just... decorating would add it to this list.  So not much would change really.

packages/validator-bonds-cli-core/src/commands/show.ts
ochaloup: Not fixing

packages/validator-bonds-sdk/src/web3.js/voteAccount.ts
ochaloup: false positive

ochaloup:
@kronael ok, I updated the PR once again to address the points from Copilot and Claude. Part of the points are relevant, so I decided to apply the fixes. Once again, please re-review.

kronael (approved):

PR #432, opened 2026-05-25
[fix] Fix dynamic bid fee calc [trivial]

kronael:
Merge only after #428, so that the target is "main".


ochaloup (dismissed):

ochaloup (dismissed):

kronael:
Ok... its ... all stuck.  Will try to push a new commit to unstuck.

ochaloup (approved):

PR #433, opened 2026-05-25
[buildkite] Add a fee report to the pipeline [GEN-7676]

kronael:
Add a script to calculate and show a fee summary table in the pipeline during processing so we can easily see how its doing.


scripts/fee-annotation.ts
```diff
@@ -0,0 +1,145 @@
...
+    0,
+  )
+  const ncap = bids.filter(
+    b =>
+      parseFloat(b.details.total_marinade_stakers_rewards) > 0 &&
+      (b.details.marinade_fee_claim ?? 0) + (b.details.dao_fee_claim ?? 0) <
+        ((parseFloat(b.details.total_marinade_stakers_rewards) * maxFeeBps) /
+          10000) *
+          0.9999,
+  ).length
```
ochaloup: I'm not sure about this, as I'm not sure if the activating stake is the real issue here. Claude argues that the SSR cap calculation is incorrect in some cases.

> SSR-cap counter overcounts when validators have activating stake - the filter compares `marinade_fee_claim + dao_fee_claim` from the Bidding row alone to `gross * max_fee_bps`, but for validators with activating stake, part of the fee is moved into a separate `PriorityFee` settlement (see `bidding.rs`). In any epoch with activating stake, every such validator is incorrectly counted as SSR-capped.

.buildkite/prepare-bid-distribution.yml
```diff
@@ -215,6 +215,17 @@ steps:
...
+  - label: ":moneybag: Fee analysis"
+    allow_dependency_failure: true
+    commands:
+    - buildkite-agent artifact download --include-retried-jobs bid-distribution-settlements.json .
+    - pnpm install
+    - |
+      pnpm exec tsx scripts/fee-annotation.ts \
+        "./bid-distribution-settlements.json" \
+        "./settlement-config.yaml" \
+        | buildkite-agent annotate --priority 6 --style info --context "fee-analysis"
```
ochaloup, line 228: Probably add `set -o pipefail` to get the error propagated? Does `fee-annotation.ts` handle the failure somehow in a way that does not break the `annotate` command?

package.json
```diff
@@ -48,8 +48,12 @@
     "prettier": "3.6.2",
     "ts-jest": "29.1.4",
     "ts-node": "10.9.2",
+    "tsx": "^4.22.3",
     "typescript": "5.4.5"
   },
+  "dependencies": {
+    "yaml": "^2.8.0"
```
ochaloup, line 53: The root `package.json` should not have any `dependencies`. This belongs to `devDependencies` I believe.

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,145 @@
+#!/usr/bin/env bun
```
ochaloup: I'm not sure about `bun` shebang here. Is it correct to have this here - while I'm not familiar with `bun` to be clear.

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,145 @@
...
+
+import { parse } from 'yaml'
+
+type Settlement = {
+  reason: string
+  details: {
+    total_marinade_active_stake: number
+    total_marinade_stakers_rewards: string
+    marinade_fee_claim: number | null
+    dao_fee_claim: number | null
```
ochaloup: As the other values here are non-null, is it necessary to allow `null` here? It seems Rust always generates some value for these fields.

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,145 @@
...
+/* eslint-disable n/no-process-exit */
+
+import { readFileSync } from 'node:fs'
+
+import { parse } from 'yaml'
+
+type Settlement = {
+  reason: string
+  details: {
+    total_marinade_active_stake: number
```
ochaloup, line 11: Maybe to use BigInt ?

ochaloup (changes requested):
I would like ask to change the `package.json`. Other points are notes that can be declined if considered not valuable.

scripts/fee-annotation.ts
kronael: Possibly yes.  Will investigate this.  Though it is not ideally preserved in the data now so would rather first fix the logging...

.buildkite/prepare-bid-distribution.yml
kronael: Not sure.  Will add pipefail.

scripts/fee-annotation.ts
kronael: It will only be used if you run it from the command line.  I can remove it.  I did have a conversation about this with Claude, but other scripts have shebangs, but then need to be invoked with pnpm anyway... so I kind of mostly ignored it as not-applicable worry.

ochaloup (commented):
Nice, but please fix the linting issue of updating the pnpm lock.

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,144 @@
...
+        break
+      }
+    } catch {}
+    if (i < 2)
+      await new Promise<void>(r => {
+        setTimeout(r, 2000)
+      })
+  }
+  if (ssrFeed === null)
+    process.stderr.write('warn: APY feed unavailable, using fallback epy=182\n')
```
ochaloup: Nitpick, non-blocking: It would be nice to define the value `182` as a constant with a comment explaining why `182`, and then use that constant throughout the code.
Btw., when checking this with Claude, isn't it from a formula `SECONDS_PER_YEAR / (0.4 * 432000) = 182.625`?

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  const cur = ssrFeed?.epochs.find(e => e.epoch === epoch)
+  const prev = ssrFeed?.epochs.find(e => e.epoch === epoch - 1)
+  if (ssrFeed !== null && (!cur || !prev)) {
+    process.stderr.write(
+      `warn: epoch ${epoch} or ${epoch - 1} not in APY feed, using fallback epy=182\n`,
+    )
+  }
+  const epy = cur && prev ? 31557600 / (cur.time - prev.time) : 182
```
ochaloup: nitpick, nonblocking: It would be nice to use a common calculation from typescript-common` or having this as a constant named `SECONDS_PER_YEAR`

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  const pmpeGross = (gross / stake) * 1000
+  const pmpeAdj = ((gross - fees) / stake) * 1000
+  const pmpeMax = ((gross * (1 - maxFeeBps / 10000)) / stake) * 1000
+  const feesSol = fees / 1e9
+  const feesFull = (gross * maxFeeBps) / 10000 / 1e9
+  const apyFor = (p: number) =>
```
ochaloup, line 144: I would like this to be part of `typescript-common`, reusable, and used as a well-defined function.

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,144 @@
...
+    if (i < 2)
+      await new Promise<void>(r => {
+        setTimeout(r, 2000)
+      })
+  }
+  if (ssrFeed === null)
+    process.stderr.write('warn: APY feed unavailable, using fallback epy=182\n')
+
+  const cur = ssrFeed?.epochs.find(e => e.epoch === epoch)
+  const prev = ssrFeed?.epochs.find(e => e.epoch === epoch - 1)
```
ochaloup: could be `cur.time === prev.time → epy = Infinity` ?

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,144 @@
...
+  )
+  // fees: all settlement types — PriorityFee carries dao_fee_claim even with no staking rewards
+  const fees = settlements.reduce(
+    (s, b) =>
+      s +
+      (b.details?.marinade_fee_claim ?? 0) +
+      (b.details?.dao_fee_claim ?? 0),
+    0,
+  )
+  const ncap = bids.filter(
```
ochaloup: Something that I don't know, but Claude points down

> The `Bidding` settlement records only the *bidding portion* of the fee. For validators with activating stake, `generate_bid_settlements` splits the fee into `Bidding` (`marinade_fee_for_bidding` / `dao_fee_for_bidding`) and a separate `PriorityFee` settlement (`marinade_fee_for_priority` / `dao_fee_for_priority`) via `activating_fee_fraction` (`bidding.rs`).
> But `total_marinade_stakers_rewards` on the Bidding row is `active_stakers_rewards + activating_bid_claim` (`bidding.rs`) — it includes the activating component. So the filter compares "bidding-only fee" against "full reward × maxFeeBps", and every validator with any activating stake is counted as SSR-capped even when it paid the full fee.

eslint.config.cjs
```diff
@@ -20,6 +21,13 @@ module.exports = [
       'no-await-in-loop': 'off',
     },
   },
+  {
+    files: ['scripts/**'],
+    plugins: { import: eslintPluginImport },
+    rules: {
+      'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
```
ochaloup: I think this is only used to avoid errors in fee-annotations, so I would consider adding an ESLint ignore rule only for that particular file rather than adding this general rule here.

If we decide it's needed, then at least add `eslint-plugin-import` to the root `package.json` devDependencies.

ochaloup (commented):
Few comments. If you consider them non-blocking, feel free to resolve them and ask for a re-review. I can approve it as is.

I’d like us to make better use of `typescript-common` for generic logic instead of reimplementing it across projects, especially for core calculations like APY. We can ship this without those changes, but it would be nice to at least create a task listing the areas that should be generalized. Either you or I could then move things in that direction during future refactoring.

kronael:
I’d like us to make better use of typescript-common for generic logic instead of reimplementing it across projects, especially for core calculations like APY. We can ship this without those changes, but it would be nice to at least create a task listing the areas that should be generalized. Either you or I could then move things in that direction during future refactoring. 

=> Yes this would be nice.  How to accomplish it here?  Why not have an api that does that rather than a library?

eslint.config.cjs
kronael: I have no idea.  It was complaining about the change you proposed earlier.  I don't mind either and don't really understand why TypeScript is such a pain to setup.  Why was it problematic to having it as dependencies.  Or if then, each of the scripts should not be its own package, to facilitate something like "uvx" allows, that is having an environment for every script.

scripts/fee-annotation.ts
kronael: Thank you for the coment.  If this is so, its we'll defined to be infinity, so ... I guess its ok to treat it that way.

scripts/fee-annotation.ts
kronael: Am thinking also on how to make the data team be able to use these...  Can we make it even into an api?

scripts/fee-annotation.ts
kronael: I agree yes.  Also fallback should not be needed to be there I guess...

scripts/fee-annotation.ts
kronael: Should not be there at all.

scripts/fee-annotation.ts
kronael: Or maybe even have a way so that we could define the functions across domains, in a single language and have them as modules elsewhere... also in rust.

eslint.config.cjs
kronael: I did add it to the script.

scripts/fee-annotation.ts
```diff
@@ -0,0 +1,164 @@
...
+  )
+  // Sum fees across all settlement types per validator (Bidding + PriorityFee).
+  // Bidding row has only the bidding-portion fee; PriorityFee carries the activating portion.
+  const feesByValidator = new Map<string, number>()
+  for (const s of settlements) {
+    if (!s.details) continue
+    const prev = feesByValidator.get(s.vote_account) ?? 0
+    feesByValidator.set(
+      s.vote_account,
+      prev + s.details.marinade_fee_claim + s.details.dao_fee_claim,
```
ochaloup:
```suggestion
      prev + (s.details.marinade_fee_claim ?? 0) + (s.details.dao_fee_claim ?? 0),
```

ochaloup (dismissed):
Just a note

kronael:
I did start with the transfer ... I have some specs for the common migration of the functions https://github.com/marinade-finance/typescript-common/pull/91.

ochaloup:
I would like to sync here that I'm not going to do the final review until the changes in `typescript-common` are finalized and published. If this PR needs to be reviewed now as it is, I can approve it in its current state.

ochaloup (approved):

PR #434, opened 2026-05-26
Bonds/Settlements refactor, fixes, consistency [GEN-7694]

ochaloup:
Heavy refactoring.

Idea: better modularized functionality and less duplication in the code.

Parts:

* bonds pipeline stops using the scoring API and loads data directly from JSON from GS, as it does for all other data like stakes, validators, etc. Only the DS SAM results were not loaded from there for some reason.
* instead of warp, using axum for the Rust API, as we do in other services. This also simplifies the rate limit code.
* introducing metrics correctly to the API
* flattening the generated JSONs to make them better structured. No big changes, but a small structural change that is not fully compatible. The API GET call should still generate the old format to not break consumers.
* big refactor of pipeline bid creation processing — decomposing the monolith into phase functions and using typed structs.
* dropping the stake-account payload from the output, as it was very clumsy without a well-defined outcome functionality. With some of the penalties starting to use `activating` stake, we started to provide pretty misleading information. The data was not used for any later pipeline steps anyway (never serialized into merkle/settlement JSON), so removing it makes good sense.
* some fixes like: merge key now includes claim kind and duplicate stake-account lamports are summed on merge
   * when multiple payouts collapse into one `(withdraw_authority, stake_authority)` StakerPayout claim, we sum `claim_amount`, `active_stake`, and each per-account `stake_accounts` balance together (map values are summed per key, not first-write-wins). Because every payout record already satisfies `active_stake == sum(stake_accounts.values())` and summation is linear, summing all three fields preserves that invariant on the merged claim, keeping the reported stake internally consistent.

Related PRs: 

- https://github.com/marinade-finance/stakes-etl/pull/66
- https://github.com/marinade-finance/ops-infra/pull/2201

## Summary by CodeRabbit

* **New Features**
  * Settlement generation can now use ds-sam auction `results.json` for SAM inputs and pipelines generate settlements from DS outputs.
  * Settlement claims are now polymorphic via a `kind` discriminator (staker payouts, fee deposits, markers).
  * API server now uses Axum, with dual listeners and new `/metrics`, `/healthz`, and `/readyz` endpoints.

* **Bug Fixes**
  * Improved legacy-format detection and clearer SAM configuration failures.
  * Error responses no longer leak internal details.

* **Documentation**
  * Updated bid/institutional distribution pipeline and CLI examples for the new inputs/outputs.

* **Tests**
  * Added API black-box HTTP tests and BigInt-safe/lossless JSON parsing tests for sanity tooling.


settlement-distributions/settlement-common/src/settlement_details.rs
```diff
@@ -0,0 +1,128 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(tag = "kind")]
+pub enum SettlementDetails {
+    Bidding(Box<BidSettlementDetails>),
+    PriorityFee(PriorityFeeSettlementDetails),
```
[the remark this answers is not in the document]
ochaloup: > serde's #[serde(tag = "kind")] is internal tagging. For a newtype variant whose inner value serializes as a map/struct, serde injects the tag key directly into that map → flat output. BidSettlementDetails is a struct, and Box<T> serializes transparently (forwards to T), so the inner value is a map → tag gets injected. The shape Copilot described — a "Bidding" sub-object — is what external tagging produces (the default when you omit tag). You cannot get both "kind" and a "Bidding" wrapper simultaneously

ochaloup:
@kronael I am refactoring and also fixing the bonds a bit. Could you review it, please?

api/src/routes.rs
```diff
@@ -0,0 +1,121 @@
...
+        .with_state(context)
+}
+
+/// Per-IP (`cf-connecting-ip`) rate limit, burst = rps. Built inline so
+/// tower_governor's multi-parameter layer type stays fully inferred against
+/// the axum router. `period` is the refill interval (rate⁻¹), not the rate.
+pub fn with_public_rate_limit(router: Router) -> Router {
+    let config = GovernorConfigBuilder::default()
+        .key_extractor(CfConnectingIpKeyExtractor)
+        .period(Duration::from_nanos(
```
kronael, line 67: 🤖 claude: **Rate-limiter keyed state never GC'd — unbounded per-IP memory growth**

The old warp implementation had a `spawn_limiter_gc()` task that called `governor_limiter.retain_recent()` on an interval. Its module doc said *"Without it, the map grows with every unique observed IP until process restart."* That task was dropped in the axum migration.

`tower_governor` 0.8 does **not** evict stale keys on its own — its own README requires spawning a cleanup task (see `tower_governor-0.8.0/README.md`). On a public Cloudflare-fronted API the per-IP `DashMap` will grow without bound until pod restart.

Fix: clone the limiter before consuming config and spawn the same interval task:
```rust
let config = GovernorConfigBuilder::default()
    // ...
    .finish()
    .expect("valid public rate-limit config");
let limiter = config.limiter().clone();
tokio::spawn(async move {
    loop {
        tokio::time::sleep(Duration::from_secs(60)).await;
        limiter.retain_recent();
    }
});
router.layer(GovernorLayer::new(config))
```

settlement-distributions/settlement-common/src/settlement_collection.rs
```diff
@@ -14,12 +14,91 @@ pub struct SettlementClaim {
     pub withdraw_authority: Pubkey,
```
kronael, line 16: 🤖 claude: **Settlement JSON format break is total, not just flattening**

The PR description presents this as a minor "flattening" change, but any consumer that reads settlement JSON (old or new) with the new or old Rust structs hits three separate breaks:

1. `"kind"` — newly required field on every claim; old files fail with `missing field 'kind'`
2. `"details"` — new internally-tagged enum; old `CommissionDetails` shape no longer deserialises
3. `"meta": { "funder": ... }` → `"funder": ...` — top-level rename

In-repo consumers were all updated (confirmed). But external consumers reading from GCS (`<epoch>/bid-distribution-settlements.json`) will break silently or loudly on the new epoch's files, and any new tooling reading old epochs will fail on (1). Deserves an explicit format-version bump note in the PR / release notes so operators know this is a hard cut — not something that can be rolled back mid-epoch without regenerating settlement files.

settlement-distributions/bid-distribution/src/generators/tests.rs
```diff
@@ -1406,22 +1419,512 @@ fn test_activating_bid_charge_distributed_to_activating_stakers() {
...
+    let staker_total = bidding
+        .claims
+        .iter()
+        .filter(|c| {
+            c.stake_authority != TEST_PUBKEY_MARINADE && c.stake_authority != TEST_PUBKEY_DAO
+        })
+        .map(|c| c.claim_amount)
+        .sum::<u64>();
+    assert_eq!(
+        staker_total, 0,
```
kronael, line 1870: 🤖 claude: **`staker_total == 0` assertion is vacuous**

The fixture for this test has exactly one staker — `TEST_PUBKEY_MARINADE` — which is also the fee withdraw authority. The iterator filter at line ~1840 excludes `stake_authority == TEST_PUBKEY_MARINADE`, so the iterator is **always empty by construction**. `staker_total` will always be 0 regardless of whether the fee logic is correct.

The fee-amount assertions above (lines ~1863–1865) are real and do catch regressions, but the staker-total assertion provides no coverage.

Fix: filter by claim variant instead of by authority:
```rust
let staker_total = bidding
    .iter()
    .filter(|s| matches!(s.claim_detail, ClaimDetail::StakerPayout { .. }))
    .map(|s| s.claim_amount)
    .sum::<u64>();
assert_eq!(staker_total, 0, "all rewards go to distributor");
```

kronael (commented):
🤖 claude: **`/protected-events` backward compat depends on the external stakes-ETL still writing a `meta` BQ column**

`api/src/repositories/protected_events.rs:24` queries `mainnet_beta_stakes.psr_settlements` for a `meta` column and `parse_row` hard-requires it — rows with a missing `meta` are silently skipped (logged at WARN). The new settlement JSON format removes the `meta` wrapper (replaces `{ "funder": ... }` with top-level `funder`). If the [stakes-etl PR #66](https://github.com/marinade-finance/stakes-etl/pull/66) writes the settlement's `funder` field to the `meta` column in a backward-compatible way, this is fine. If not, every post-deploy epoch's PSR settlement records will be silently dropped and the `/protected-events` endpoint freezes at the last pre-deploy epoch.

This is an implicit cross-repo contract that isn't validated by any test here. Worth explicitly verifying against the ETL schema before merging, or making `parse_row` accept a funder-shaped row as a fallback.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -55,644 +58,576 @@ impl fmt::Display for ResultSettlementClaims {
...
+}
+
+fn fraction_or_fallback(numerator: Decimal, denominator: Decimal, fallback: Decimal) -> Decimal {
+    if denominator > Decimal::ZERO {
+        numerator / denominator
+    } else {
+        fallback
+    }
+}
+
```
kronael, line 115: 🤖 claude: **Fee bisection skewed for activating-only validators**

For a validator that has a `PriorityFee` settlement but no `Bidding` settlement (line 112 guard), the totals function adds `activating_bid_claim` to `totals.rewards` (line 115) but only `total_marinade_active_stake` to `totals.stake` (line 113). For a pure-activating validator (zero active stake, non-zero activating stake), `total_marinade_active_stake == 0`, so `totals.stake` gets nothing while `totals.rewards` picks up the activating claim.

This inflates the global post-fee PMPE numerator without adding to the denominator, making the fee bisection see higher yield than it should for the epoch — potentially selecting a higher `max_fee_bps` than is warranted.

No existing test exercises the bisection feasibility path for a pure-activating validator (all activating tests in `tests.rs` use `create_test_fee_config(0,0)` or `min==max`, bypassing bisection).

Fix: either add `Decimal::from(value.total_marinade_activating_stake)` to `totals.stake` in the `PriorityFee` branch, or add a targeted test that verifies the bisection outcome with an activating-only validator.

settlement-distributions/merkle-generator/src/merkle_generator.rs
```diff
@@ -116,7 +116,7 @@ pub fn generate_merkle_tree_collection(
     for source in &sources {
         for settlement in &source.collection.settlements {
             grouped_settlements
-                .entry((settlement.vote_account, settlement.meta.funder.clone()))
+                .entry((settlement.vote_account, settlement.funder.clone()))
```
kronael, line 119: 🤖 claude: **`merge_claims` ignores `ClaimDetail` kind — inconsistent with institutional hardening**

`merge_claims` (lines 81–93, not changed in this PR) keys only on `(withdraw_authority, stake_authority)` and sums `claim_amount`, discarding the `ClaimDetail` variant. The institutional `settlement_generator.rs` was deliberately hardened in this PR (lines 126–131) to keep `StakerPayout` and `FeeDeposit` claims separate even when they share the same authority pair.

The final cross-settlement merge in `merkle_generator.rs` doesn't carry that guard. If a fee-deposit authority pair ever appeared as a staker payout authority (or vice versa), the two kinds would silently coalesce into a single tree node with summed amounts — the exact issue the institutional fix was designed to prevent.

Not breaking today (Marinade/DAO fee-deposit authorities are always distinct from staker authority pairs in practice), but the invariant is asserted nowhere. Consider adding a `debug_assert!` or a pre-merge collision check analogous to the institutional one, so the assumption is explicit and auditable.

settlement-distributions/settlement-common/src/settlement_collection.rs
ochaloup: Acknowledged — agreed this is an intentional hard cut, not a recoverable flattening. The settlement JSON is forward-only: there is no dual-shape read path, and all in-repo consumers are already updated for the new `kind` / `details` / top-level `funder` shape.

I'll add an explicit format-version / hard-cut note to the PR description and release notes so operators know this cannot be rolled back mid-epoch without regenerating the settlement files. This pairs with the stakes-ETL deploy-ordering note (the BQ `meta` column is reconstructed from the flat `funder` there). Resolving as documented forward-only debt.

settlement-distributions/bid-distribution/src/generators/tests.rs
ochaloup: Addressed. Both `test_activating_fee_fraction_falls_back_*` assertions now filter on `matches!(c.detail, ClaimDetail::StakerPayout { .. })` instead of the authority pair, so the staker-total check no longer passes vacuously (the sole staker authority was previously filtered out by construction). `cargo test -p bid-distribution` and clippy are green.

settlement-distributions/bid-distribution/src/generators/bidding.rs
ochaloup: Declined as a code change — the asymmetry is intentional and consistent: the Bidding branch and `split_distributor_fee` both fold `activating_bid_claim` into the yield numerator while normalizing by active+redelegation stake only, so adding activating stake to the denominator here alone would skew activating-only validators vs mixed ones (and break zero-behavior-change). I closed the coverage gap instead with `test_calculate_bid_settlement_totals_priority_fallback_excludes_activating_stake`, which pins this behavior so any future change to the denominator is a conscious one.

settlement-distributions/merkle-generator/src/merkle_generator.rs
ochaloup: Declined. The merkle leaf is keyed only on `(withdraw_authority, stake_authority, claim)` with no claim kind (`merkle-tree/src/psr_claim.rs`), so a same-pair StakerPayout+FeeDeposit collision is correctly resolved by summing into one leaf — the recipient is owed the total, and the on-chain destination is identical regardless of kind. A hard `Err` would reject a benign, well-defined case (and a `debug_assert!` is stripped from the release pipeline anyway); the institutional check guards different state — the StakerPayout `active_stake`/`stake_accounts` metadata during intra-settlement accumulation — which no longer exists at the merkle stage.

ochaloup:
@kronael I adjusted the code based on your review. Sorry for the force push, it happened :smile:

All comments were either addressed or replied to.

Backwards compatibility is not fully considered here. It’s a bit of a hard cut to have the better-aligned `reason`/`funder` JSON format. If you think we should keep compatibility — because of analytics or something else — then I would need to rethink the solution a bit. One option could be to have some script that makes the JSON style cross-compatible.

settlement-distributions/bid-distribution/src/bin/cli.rs
```diff
@@ -40,10 +40,14 @@ struct Args {
     settlement_config: String,
 
     // ===== SAM-specific inputs (optional) =====
-    /// SAM scoring meta collection JSON file
+    /// SAM scoring meta collection JSON file (scoring-API `/scores/sam` shape)
     #[arg(long, env)]
     sam_meta_collection: Option<String>,
 
+    /// ds-sam auction `results.json` (reads `auctionData.validators`); alternative to --sam-meta-collection
```
kronael, line 51: Do we need to really support both?

settlement-pipelines/src/settlement_data.rs
```diff
@@ -282,7 +282,7 @@ pub fn parse_settlements_from_json(
                     tree_nodes: merkle_tree.tree_nodes.clone(),
                     max_total_claim_sum: merkle_tree.max_total_claim_sum,
                     max_total_claim: merkle_tree.max_total_claims as u64,
-                    funder: SettlementFunderType::new(&settlement.meta.funder),
+                    funder: SettlementFunderType::new(&settlement.funder),
```
kronael, line 285: Love it when its less nested.

settlement-distributions/settlement-common/src/settlement_details.rs
```diff
@@ -0,0 +1,128 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Deserialize, Serialize)]
+#[serde(tag = "kind")]
```
kronael, line 4: If I were to choose I'd just choose outer-tagged.  But I guess we need to remain compatible here.  Why: Because then it leads to less code in typescript and type-checks on the instance can be done with less code, less knowledge about the "class-tag" name.

settlement-distributions/settlement-common/src/settlement_details.rs
```diff
@@ -0,0 +1,128 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct BondRiskFeeDetails {
+    pub total_marinade_active_stake: u64,
+    pub effective_sam_marinade_active_stake: u64,
+    pub bond_risk_fee_sol: String,
+    pub stakers_bond_risk_fee_claim: u64,
+}
+
+#[cfg(test)]
+mod tests {
```
kronael: Would just suggest here that we could use the opportunity and move the tests off-of the main file.

settlement-distributions/settlement-common/src/settlement_collection.rs
```diff
@@ -61,6 +140,9 @@ pub enum SettlementFunder {
     Marinade,
 }
 
+// Retained for the settlement-config YAML input and the protected-events API response,
```
kronael: I dont find this comment useful.

settlement-distributions/settlement-common/src/settlement_collection.rs
```diff
@@ -14,12 +14,91 @@ pub struct SettlementClaim {
...
+        stake_accounts: HashMap<Pubkey, u64>,
+    },
+    FeeDeposit,
+    // Zero-amount placeholder distinguishing Marinade- vs ValidatorBond-funded merkle
+    // roots when claim amounts would otherwise coincide. Not a real payout.
+    Marker,
+}
+
+impl SettlementClaim {
+    pub fn staker_payout(
```
kronael, line 43: I wonder, why not put these there verbatim at the construction sites.  Rust has the unfortunate feature of not supporting keyword params...

settlement-distributions/bid-distribution/tests/fixtures/golden_settlement.json
kronael, line 1: golden ... so Claude :D

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -55,644 +58,576 @@ impl fmt::Display for ResultSettlementClaims {
...
+        ..
+    }) = &validator.values
+    {
+        let staker_inflation_rewards =
+            mr.inflation * (Decimal::ONE - commissions.inflation_commission_dec);
+        let staker_mev_rewards = mr.mev * (Decimal::ONE - commissions.mev_commission_dec);
+        let staker_block_rewards =
+            mr.block * (Decimal::ONE - commissions.block_rewards_commission_dec);
+        let staker_bid_rewards = result_claims.static_bid_claim;
+        let total = staker_inflation_rewards
```
kronael: Why not "active_total"?

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -55,644 +58,576 @@ impl fmt::Display for ResultSettlementClaims {
...
+        .min(result_claims.sum())
+        .to_u64()
+        .ok_or_else(|| anyhow!("Failed to_u64 for distributor_fee_claim"))?;
+
+    // minimum is 0 when distributor fee is of amount of total (stakers get nothing)
+    let settlement_claim_sum = result_claims.sum_u64()?;
+    let stakers_total_claim = settlement_claim_sum.saturating_sub(distributor_fee_claim);
+    // Split stakers_total_claim between active stakers (earned rewards + static bid)
+    // and activating stakers (activating charge), proportional to each pool's share.
+    let activating_fraction = fraction_or_fallback(
```
kronael: Just an idea... what if this was called like:

check_zero_divide(a, b).unwrap_or(Decimal::Zero)

Would it preclude some of the uses?  It might be "less esoteric" that way on what is happening.  More composable.

kronael (dismissed):
Mostly nits.

settlement-distributions/settlement-common/src/settlement_details.rs
ochaloup: Internal tag is the deliberate flat-output choice; outer-tag would re-break JSON. I'm sorry.

settlement-distributions/settlement-common/src/settlement_details.rs
ochaloup: I like the idea. Let me investigate.

settlement-distributions/settlement-common/src/settlement_collection.rs
ochaloup: constructors encapsulate the detail-variant invariant; literals would duplicate it everywhere

settlement-distributions/bid-distribution/src/generators/bidding.rs
ochaloup: Ok. A good idea.

settlement-distributions/bid-distribution/src/bin/cli.rs
ochaloup: ~~Agree here, probably not needed. I was thinking it could be useful for loading data from past runs. But when you ask directly...~~

I investigated, and it's better to have both. There is tooling around testing and simulation that works with both, so it's good to support both.

ochaloup:
@kronael I rebased, updated the code and added the comments. Can you please re-review :pray:

ochaloup:
I updated the PR based on Copilot's findings. Still ready for review. Thinking to do some touches athttps://github.com/marinade-finance/stakes-etl/pull/66 as per https://github.com/marinade-finance/validator-bonds/pull/434#pullrequestreview-4478037394

settlement-distributions/settlement-common/src/settlement_details.rs
ochaloup: 🤖 claude: Done in commit 79571cf — moved the inline `#[cfg(test)] mod tests` out of `settlement_details.rs` into a crate-level integration test `settlement-common/tests/settlement_details.rs` (both tests pass there).

settlement-distributions/settlement-common/src/settlement_collection.rs
ochaloup: 🤖 claude: Removed in commit 79571cf. (Kept the neighboring `Marker` comment, which documents a non-obvious why.)

settlement-distributions/bid-distribution/src/generators/bidding.rs
ochaloup: 🤖 claude: Renamed the local `total` to `active_total` in both branches of `staker_rewards_breakdown` (commit 79571cf).

settlement-distributions/bid-distribution/src/generators/bidding.rs
ochaloup: 🤖 claude: Adopted in commit 79571cf — `fraction_or_fallback` is now `checked_fraction(n, d) -> Option<Decimal>` with `.unwrap_or(...)` at the call sites. Byte-identical settlement output verified against main for epochs 978-980.

api/tests/http_behavior.rs
```diff
@@ -0,0 +1,242 @@
...
+    format!("http://{addr}")
+}
+
+/// reqwest client with redirects/compression off so raw headers are observable.
+fn client() -> reqwest::Client {
+    reqwest::Client::builder()
+        .redirect(reqwest::redirect::Policy::none())
+        .build()
+        .unwrap()
+}
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: No change needed — reqwest gzip is an opt-in feature, not enabled here. `Cargo.toml` pins `reqwest = "0.11.22"` with default features only, and `api/Cargo.toml` documents reqwest default features (no gzip). The test client therefore neither sends `Accept-Encoding` nor auto-decompresses, so the gzip-negotiation assertions stay valid. Resolving as a false positive.

scripts/regression-test-settlements.sh
```diff
@@ -742,6 +747,48 @@ process_epoch() {
...
+  # =====================================================================
+  # When a local ds-sam results.json exists for this epoch, verify the production
+  # input path (--sam-results-collection, mapped by into_validator_sam_metas) yields
+  # the SAME settlements as the scoring-API --sam-meta-collection path compared above.
+  # Closes the gap where regression never exercised the mapping production runs.
+  local sam_results_status="SKIP"
+  local results_json=""
+  if compgen -G "$DS_SAM_AUCTIONS_DIR/$epoch."*"/outputs/results.json" > /dev/null 2>&1; then
+    results_json=$(ls -d "$DS_SAM_AUCTIONS_DIR/$epoch."*"/outputs/results.json" 2>/dev/null | head -1)
+  fi
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed in commit 687384a — `results_json` now uses `sort -V | tail -1` instead of `head -1`, so the latest `<epoch>.<slot>` is selected deterministically (matching the README guidance to use the latest slot).

packages/validator-bonds-sanity-check/src/dtoSettlements.ts
```diff
@@ -198,11 +216,41 @@ export class SettlementsDto {
...
+    first?.funder === undefined && first?.meta?.funder !== undefined
+  const legacyClaim =
+    Array.isArray(first?.claims) &&
+    first.claims.length > 0 &&
+    first.claims.every((claim: any) => claim?.kind === undefined)
+  if (legacyFunder || legacyClaim) {
+    throw CliCommandError.instance(
+      `'settlements' data at path '${path}' is in the legacy pre-refactor format (missing top-level 'funder' and/or per-claim 'kind'). Legacy files are not supported — regenerate this epoch with the current pipeline to produce the new format.`,
+    )
+  }
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed in commit 687384a — `assertNotLegacyFormat` now scans all settlements (`settlements.some(isLegacy)`) instead of only `settlements[0]`, and the error message omits the path suffix when no path is provided (no more `at path undefined`).

ochaloup:
@kronael would you have a while to review?

packages/validator-bonds-sanity-check/src/dtoSettlements.ts
```diff
@@ -198,11 +216,44 @@ export class SettlementsDto {
...
+    return
+  }
+  const isLegacy = (settlement: any): boolean => {
+    const legacyFunder =
+      settlement?.funder === undefined && settlement?.meta?.funder !== undefined
+    const legacyClaim =
+      Array.isArray(settlement?.claims) &&
+      settlement.claims.length > 0 &&
+      settlement.claims.every((claim: any) => claim?.kind === undefined)
+    return legacyFunder || legacyClaim
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Addressed — the per-claim legacy check now uses `.some(claim => claim?.kind === undefined)` instead of `.every(...)`, so a settlement with any kind-less claim (partial/corrupt file) is flagged with the clear error rather than slipping to a generic class-validator failure. No false positives: new-format claims all carry `kind`.

api/tests/http_behavior.rs
```diff
@@ -0,0 +1,242 @@
...
+    tokio::spawn(async move {
+        axum::serve(
+            listener,
+            app.into_make_service_with_connect_info::<SocketAddr>(),
+        )
+        .await
+        .unwrap();
+    });
+    // Give the spawned server a moment to start accepting connections.
+    tokio::time::sleep(Duration::from_millis(50)).await;
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Addressed — replaced the fixed 50ms sleep with a `wait_until_accepting(addr)` helper that retries `TcpStream::connect` (up to ~1s) until the server accepts, so the test no longer depends on a fixed startup delay.

api/tests/http_behavior.rs
```diff
@@ -0,0 +1,242 @@
...
+    let listener = tokio::net::TcpListener::bind(("127.0.0.1", 0))
+        .await
+        .unwrap();
+    let addr = listener.local_addr().unwrap();
+    tokio::spawn(async move {
+        axum::serve(listener, app.into_make_service())
+            .await
+            .unwrap();
+    });
+    tokio::time::sleep(Duration::from_millis(50)).await;
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Addressed — both spawn helpers now share the same `wait_until_accepting(addr)` connect-retry helper instead of a fixed 50ms sleep.

ochaloup:
Static analysis fixed, now even more greenly ready for review @kronael

api/tests/http_behavior.rs
kronael, line 1: ❤️

scripts/regression-test-settlements.sh
```diff
@@ -742,6 +747,48 @@ process_epoch() {
     fi
   fi
 
+  # =====================================================================
+  # SAM INPUT-PATH EQUIVALENCE (production uses --sam-results-collection)
+  # =====================================================================
+  # When a local ds-sam results.json exists for this epoch, verify the production
+  # input path (--sam-results-collection, mapped by into_validator_sam_metas) yields
+  # the SAME settlements as the scoring-API --sam-meta-collection path compared above.
```
kronael, line 779: Could we just.... only support the "on-git" path and drop ds-scoring support and data format?

settlement-distributions/bid-distribution/src/generators/fee_optimizer.rs
```diff
@@ -0,0 +1,229 @@
...
+pub fn calculate_bid_settlement_totals(settlements: &[Settlement]) -> BidSettlementTotals {
+    let mut totals = BidSettlementTotals::default();
+    let bidding_votes: HashSet<Pubkey> = settlements
+        .iter()
+        .filter(|s| matches!(s.reason, SettlementReason::Bidding))
+        .map(|s| s.vote_account)
+        .collect();
+    for settlement in settlements {
+        match (&settlement.reason, &settlement.details) {
+            (SettlementReason::Bidding, Some(SettlementDetails::Bidding(value))) => {
```
kronael: Why not match only on the "detail"?

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -505,6 +523,17 @@ fn aggregate_rewards(
         rewards_map.len()
     );
 
+    // commission_eff clamps negative realized commissions so bonds are never overcharged, but the
+    // aggregate assert above can't see per-validator excesses (they net out). Surface each offender.
+    for rewards in rewards_map.values() {
+        for (kind, commission) in rewards.negative_realized_commissions() {
```
kronael: Is it worth it defining the strangely looking function just to call it here?

kronael (dismissed):

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -55,705 +58,571 @@ impl fmt::Display for ResultSettlementClaims {
...
+fn split_distributor_fee(
+    result_claims: &ResultSettlementClaims,
+    total_marinade_stakers_rewards: Decimal,
+    totals: &StakeTotals,
+    fee_percentages: &FeePercentages,
+    target_pmpe: Decimal,
+    vote_account: &Pubkey,
+) -> anyhow::Result<FeeSplit> {
+    let effective_fee = if total_marinade_stakers_rewards > Decimal::ZERO
+        && totals.marinade_active > 0
```
kronael: 🤖 claude: **CRITICAL — activating stakers underpaid when Marinade active stake is zero.** This gate is `total_marinade_stakers_rewards > 0 && totals.marinade_active > 0`, but the yield denominator on the next line is `marinade_active + marinade_redelegation` — the gate contradicts its own formula. When a validator's Marinade stake is entirely activating/deactivating (`marinade_active == 0`, `marinade_redelegation > 0`, `rewards > 0`), this falls to `else` and charges `max_fee` instead of the lower yield-capped fee. Pre-refactor (deleted lines ~1030-1041) gated on `marinade_active + marinade_redelegation > 0` then `staker_yield_pmpe > 0`, so this state took the reduction; the refactor dropped redelegation from the gate. Reachable: a group with `active=0, activating>0, deactivating>0` isn't skipped (skip needs `active==0 && activating==0`) and has `rewards>0` via `activating_bid_claim`. Effect: distributor (Marinade/DAO) overcharges, activating stakers underpaid by `(max_fee − yield_capped_fee) × stakers_rewards`. The active==0 tests miss it (zero-fee configs or `deactivating=0`). **Fix:** gate on `totals.marinade_active + totals.marinade_redelegation > 0`.

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -87,8 +87,7 @@ impl FeeConfig {
         if let Some(value) = self.min_yield_premium_over_ssr_pmpe {
             ensure!(
                 value.abs() <= dec!(0.1),
-                "min_yield_premium_over_ssr_pmpe {} exceeds ±0.1 PMPE (~2% APY)",
-                value
+                "min_yield_premium_over_ssr_pmpe {value} exceeds ±0.1 PMPE (~2% APY)"
```
kronael, line 90: 🤖 claude: `settlement_config.rs` is declared READ-ONLY in CLAUDE.md but this changes it. The change is benign (clippy `uninlined_format_args`: `"...{}", value` → `"...{value}"`, identical behavior) — either revert to honor the rule or relax the rule's wording.

settlement-distributions/bid-distribution/src/sam_meta.rs
```diff
@@ -82,3 +82,158 @@ pub struct CommissionDetails {
...
+                    stake_priority: v.stake_priority,
+                    unstake_priority: v.unstake_priority,
+                    max_stake_wanted: v.max_stake_wanted.unwrap_or(Decimal::ZERO),
+                    effective_bid,
+                    constraints: v
+                        .last_cap_constraint
+                        .map(|c| c.constraint_type)
+                        .unwrap_or_default(),
+                    // scoring_run_id/metadata are scoring-API artifacts not present in results.json; unused by generators
+                    metadata: SamMetadata::default(),
```
kronael, line 147: 🤖 claude: This GCS `results.json` path zero-fills `metadata` (`SamMetadata::default()`) and `scoring_run_id: 0`. Harmless for in-tree generators (the regression equivalence check proves identical settlement output), but the two SAM input paths (`--sam-meta-collection` vs `--sam-results-collection`) are NOT output-identical, despite the README presenting them as drop-in alternatives. An out-of-tree consumer reading `marinadeSamTvlSol`/`scoringId` would get zeros.

settlement-distributions/bid-distribution/src/sam_meta.rs
```diff
@@ -82,3 +82,158 @@ pub struct CommissionDetails {
...
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub struct SamLastCapConstraint {
+    pub constraint_type: String,
+}
+
+impl SamAuctionResult {
+    pub fn into_validator_sam_metas(self) -> Vec<ValidatorSamMeta> {
+        let epoch = self.auction_data.epoch;
+        self.auction_data
```
kronael, line 128: 🤖 claude: (LOW, not a regression) `auctionData.validators` isn't deduped by `vote_account` here; a duplicate row would produce two settlement-claim sets over the same staker stake (≈doubled claims). Not reachable from current upstream (ds-sam is one-row-per-validator) and the old code behaved identically — worth a dedup guard for defense-in-depth only.

api/src/error.rs
```diff
@@ -1,13 +1,25 @@
...
+
+impl IntoResponse for AppError {
+    fn into_response(self) -> Response {
+        log::error!("{self:?}");
+        (
+            StatusCode::INTERNAL_SERVER_ERROR,
+            "Internal server error".to_owned(),
+        )
+            .into_response()
     }
```
kronael, line 24: 🤖 claude: The 500 body changed from warp's default rejection body to the literal `"Internal server error"` — intentional and an improvement (no leak, Capitalized), just a heads-up for any consumer that string-matched the old body. Minor: the underlying DB error is collapsed into a debug string (`{error:?}`), so connection-vs-query context is lost in logs.

settlement-distributions/bid-distribution/src/generators/tests.rs
```diff
@@ -1421,22 +1436,516 @@ fn test_activating_bid_charge_distributed_to_activating_stakers() {
...
+        Decimal::ZERO,
+        BisectMode::TargetStakerPmpe,
+    )
+    .unwrap()
+    .settlements;
+    // Stable order: sort by reason then vote_account (matches CLI).
+    settlements.sort_by_key(|s| (s.reason.to_string(), s.vote_account));
+    // details contains internal computation context; strip to keep snapshot focused on claim shape.
+    for s in &mut settlements {
+        s.details = None;
```
kronael, line 1767: 🤖 claude: The golden snapshot sets `s.details = None` before comparing, so the new `kind`-tagged `SettlementDetails` JSON serialization isn't pinned by this fixture. It's covered by the `settlement_details.rs` round-trip + per-field asserts, so not a hole — just noting the golden itself doesn't lock the `details` shape.

settlement-distributions/institutional-distribution/src/settlement_generator.rs
```diff
@@ -119,53 +98,71 @@ fn merge_payouts(
...
                 claims_count: 0,
                 claims_amount: 0,
                 claims: vec![],
                 details: None,
             });
-        settlement.claims_count += 1;
         settlement.claims_amount += payout.payout_lamports;
 
+        let incoming_is_staker = matches!(payout.kind, PayoutKind::Staker);
         if let Some(existing_claim) = settlement.claims.iter_mut().find(|claim| {
```
kronael, line 123: 🤖 claude: (Confirm intended) Adding `kind` to the JSON-level merge key here keeps a `StakerPayout` + `FeeDeposit` on the same `(withdraw, stake)` as 2 distinct claims, but `merkle_generator.rs:82` `merge_claims` still keys by authorities only and collapses them into 1 on-chain leaf (summed amount). So JSON `claims_count` can exceed the merkle leaf count on a cross-kind authority collision. Verified benign — amounts reconcile, on-chain sizing uses the merkle leaf count, and `checkMerkleTree.ts` tolerates count divergence — but worth confirming the JSON-richer / merkle-collapsed split is intended.

settlement-distributions/bid-distribution/README.md
```diff
@@ -2,19 +2,30 @@
...
+# SAM auction results from the ds-sam-pipeline GitHub repo: pick the highest
+# <slot> directory for the epoch under auctions/<epoch>.<slot>/
+curl -o "${pipeline_dir}/results.json" \
+    "https://raw.githubusercontent.com/marinade-finance/ds-sam-pipeline/main/auctions/${epoch}.<slot>/outputs/results.json"
+
+# evaluation.json: produced by ds-sam `analyze-revenues` over the auction
+# inputs/outputs and validators.json (see the "DS SAM generates PMPE bids data"
+# step in the buildkite pipeline)
 
 # SSI/SSR pmpe for the scoring epoch is fetched by the CLI from apy-api at
```
kronael, line 31: 🤖 claude: Two doc gaps: (1) the data-gathering section references `evaluation.json` / `--revenue-expectation-collection` (ds-sam `analyze-revenues`) without a command to fetch/produce it; (2) `api/README.md` is missing the "manual smoke test" that the new API integration tests (`api/tests/http_behavior.rs`) explicitly defer to for the DB-backed routes + `readyz`.

settlement-distributions/bid-distribution/src/settlement_config.rs
ochaloup: 🤖 claude: Declining — keeping the inlined `{value}` form.

To be clear, no clippy rules or config were changed in the repo: there's no `clippy.toml`, no `[lints]` table, and no clippy flags in CI (`lint:cargo` is just `cargo fmt --check && cargo clippy --release`, no `-D warnings`). This one-liner is a benign, behavior-identical `uninlined_format_args` autofix (`"...{}", value` → `"...{value}"`) that got swept in alongside the negative-realized-commission change (8c1e7d6).

`uninlined_format_args` is warn-by-default, so the inlined form is the clippy-clean one — reverting to `"...{}", value` would re-introduce the warning. We're fine taking the formatting change here rather than the revert. The `settlement_config.rs` "read-only" guideline is about not altering fee / settlement-type *semantics*, which a format-string cleanup doesn't touch. Resolving.

settlement-distributions/bid-distribution/src/generators/fee_optimizer.rs
ochaloup: 🤖 claude: Done — the match now keys on `&settlement.details` only (`Some(SettlementDetails::Bidding(..))` / `Some(SettlementDetails::PriorityFee(..))`).

The `reason` half was redundant: the detail variant is 1:1 with the reason, so the variant alone determines it. And in the only case where the two would differ — a `reason`/`details` mismatch that no constructor produces — the tuple match silently dropped the settlement into `_ => {}` instead of surfacing the inconsistency, which is worse than useless as a guard. `reason` is still used where it actually matters (the `bidding_votes` set).

I also updated `test_calculate_bid_settlement_totals_skips_non_bidding` to use a realistic non-bid settlement (a `BidTooLowPenalty` carrying its own details) instead of the previous self-contradictory fixture (`reason = BidTooLowPenalty` with Bidding details), keeping the exact same zero-assertions. All bid-distribution tests pass and the 978–980 settlement output is byte-identical. Resolving.

settlement-distributions/bid-distribution/src/rewards.rs
ochaloup: 🤖 claude: Keeping `negative_realized_commissions()` as a method.

It encapsulates the per-kind (inflation / mev / block) "realized commission exceeds the in-bond promise" detection on the rewards type and returns only the offending `(kind, commission)` pairs, so the warn loop in `aggregate_rewards` stays a flat `for (kind, commission) in … { warn!(…) }`. Inlining the three per-kind comparisons at the single call site would just move that branching into the loop for no real gain, and the method keeps the logic unit-testable in isolation. Resolving.

api/src/error.rs
ochaloup: 🤖 claude: Intentional and staying.

The opaque `"Internal server error"` 500 body is the deliberate improvement over warp's default rejection body — no internal detail is leaked to the client. The full error is still logged server-side via `log::error!("{self:?}")`, so operators retain visibility. The minor note about the DB error being folded into the `message` string is acknowledged but not worth a change here — the relevant context is captured where the error is wrapped into `AppError`. Resolving.

settlement-distributions/bid-distribution/src/generators/tests.rs
ochaloup: 🤖 claude: Intentional, not a coverage hole.

The golden snapshot strips `details` (`s.details = None`) on purpose to keep the fixture focused on claim shape and amounts. The `SettlementDetails` JSON serialization is pinned separately by the round-trip + per-field assertions in `settlement-common/tests/settlement_details.rs`, which is a more targeted test than a broad golden would be. So `details` is covered — just not by this fixture. Leaving as-is. Resolving.

settlement-distributions/bid-distribution/README.md
ochaloup: 🤖 claude: Done — both doc gaps filled, kept minimal.

- `settlement-distributions/bid-distribution/README.md`: added a one-line GCS fetch for `evaluation.json` next to the existing results.json/rewards fetches — `gcloud storage cp "gs://$bucket/$epoch/bid-psr-distribution-evaluation.json" …`.
- `api/README.md`: added a short note that `api/tests/http_behavior.rs` covers routing/middleware only, while the DB-backed routes (`/bonds/*`, `/protected-events`) and `readyz` are smoke-tested manually against the documented dev steps.

Resolving.

scripts/regression-test-settlements.sh
ochaloup: I'm not sure about it. I currently lean toward leaving it here as a working variant and a double-check option (both input formats produce identical output).

But maybe you're right that it is better to simplify. Can you express your thinking on this once more? I can remove it.

settlement-distributions/bid-distribution/src/sam_meta.rs
ochaloup: 🤖 claude: Added a hard-error dedup guard.

I placed it in `bid/src/bin/cli.rs` right after `sam_validator_metas` is assembled, so it covers **both** input paths (`--sam-meta-collection` and `--sam-results-collection` → `into_validator_sam_metas`) at the point where they converge, rather than just the results.json parser. A duplicate `vote_account` would otherwise drive two full settlement-claim sets over the same staker stake (≈doubled claims), so it now `ensure!`s uniqueness and bails listing the offending vote accounts — a loud failure rather than a silent dedup, per our strictness preference.

Not reachable from ds-sam today (one row per validator); this is defense-in-depth. 978–980 regression bid output is byte-identical and the `--sam-results` equivalence path still MATCHes. Resolving.

settlement-distributions/bid-distribution/src/sam_meta.rs
ochaloup: 🤖 claude: Verified — false positive for in-tree consumers.

Grepped every use of `metadata` / `scoring_run_id` across `bid-distribution/src`: both are only ever *written* at the two parse sites and never *read* by any generator (`SamMetadata`'s `scoring_id` / `tvl` fields are `pub(crate)` and unread too). They don't reach the settlement output either — the regression's own `--sam-results` vs `--sam-meta` equivalence check is byte-identical. The only thing that would observe the zero-fill is a hypothetical out-of-tree tool reading `ValidatorSamMeta` directly, which doesn't exist in this repo. No change needed. Resolving.

settlement-distributions/institutional-distribution/src/settlement_generator.rs
ochaloup: 🤖 claude: Confirmed correct and on-chain-neutral.

The kind-aware JSON split (keeping `StakerPayout` and `FeeDeposit` as distinct claims on the same authority pair) is intentional — it preserves the staker's stake metadata that main's authority-only merge would have coalesced away. The merkle generator still keys leaves on `(withdraw_authority, stake_authority)` only, so it sums the two into one leaf, and on-chain account sizing uses `tree_nodes.len()` (the merged leaf count), not the JSON `claims_count`.

So `claims_count` > leaf count is purely a JSON reporting nuance with zero on-chain effect: amounts reconcile, and it requires a cross-kind authority collision that doesn't occur for institutional data (staker authorities are distinct from the Marinade/DAO fee authorities). The split is new in this PR, but the 978–980 institutional output is a full MATCH vs production. Resolving.

settlement-distributions/institutional-distribution/src/settlement_generator.rs
```diff
@@ -1,20 +1,24 @@
...
 use crate::settlement_config::InstitutionalDistributionConfig;
-use settlement_common::stake_meta_index::StakeMetaIndex;
 use settlement_common::utils::sort_claims_deterministically;
 use solana_sdk::pubkey::Pubkey;
 use std::collections::HashMap;
 
+enum PayoutKind {
+    Staker,
+    FeeDeposit,
+}
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: False positive — this compiles.

`matches!(payout.kind, PayoutKind::Staker)` and `match payout.kind { PayoutKind::Staker => …, PayoutKind::FeeDeposit => … }` match on non-binding unit-variant patterns, which only read the discriminant — they do **not** move `payout.kind` (a move would require a by-value binding pattern, and there is none). `payout` is owned by the `for payout in payouts` loop, so the subsequent by-value field uses (`payout.withdrawer`, `payout.stake_amount`, …) are valid regardless of `PayoutKind` not being `Copy`. Confirmed: `cargo check -p institutional-distribution` is clean and the CLI builds and runs in the regression. Resolving.

settlement-distributions/institutional-distribution/src/settlement_generator.rs
```diff
@@ -119,53 +98,71 @@ fn merge_payouts(
...
-        settlement.claims_count += 1;
         settlement.claims_amount += payout.payout_lamports;
 
+        let incoming_is_staker = matches!(payout.kind, PayoutKind::Staker);
         if let Some(existing_claim) = settlement.claims.iter_mut().find(|claim| {
-            claim.withdraw_authority == payout.withdrawer && claim.stake_authority == payout.staker
+            let existing_is_staker = matches!(claim.detail, ClaimDetail::StakerPayout { .. });
+            claim.withdraw_authority == payout.withdrawer
+                && claim.stake_authority == payout.staker
+                && existing_is_staker == incoming_is_staker
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: False positive — this compiles.

`matches!(claim.detail, ClaimDetail::StakerPayout { .. })` uses the struct-variant pattern `{ .. }`, which binds no fields and therefore moves nothing out of the `&mut SettlementClaim` borrow — it only tests the discriminant by reference. Confirmed: `cargo check -p institutional-distribution` is clean. Resolving.

ochaloup:
@kronael, I updated the PR based on your review points. Thanks for them, those ones worked fine to investigate them.
I ask for a re-review, and we have one open discussion here.

ochaloup:
@kronael this should be ready for re-review

api/README.md
```diff
@@ -57,3 +57,7 @@ cargo run --bin api -- --postgres-url "$POSTGRES_URL" \
 # data is gzipped so we use curl --compressed
 curl -X GET --compressed "http://localhost:8000/bonds/bidding"
 ```
+
+The integration tests (`api/tests/http_behavior.rs`) cover routing/middleware only; the
```
kronael, line 61: We could have testcontainers here for these routes too. (possibly next PR or so)

api/src/metrics.rs
```diff
@@ -0,0 +1,109 @@
...
+        &["method", "path", "status"]
+    )
+    .expect("metric can be registered")
+});
+
+/// axum middleware: record request count + latency. Uses the matched route
+/// template (e.g. `/bonds/bidding`) as the `path` label to keep cardinality
+/// bounded; unmatched requests are bucketed under `unknown`.
+pub async fn track_metrics(req: Request, next: Next) -> Response {
+    let method = req.method().as_str().to_owned();
```
kronael: 🤖 claude: **Unbounded `method` label → Prometheus cardinality blowup / memory exhaustion**

`method` is taken raw, but the HTTP method is attacker-controlled and unbounded (hyper accepts any RFC token, not just GET/POST). `IntCounterVec`/`HistogramVec` retain one child series per distinct `(method, path, status)` tuple for the life of the process — no eviction (unlike the rate-limiter store you GC in `routes.rs`). `path` is correctly bounded via `MatchedPath` and `status` is finite, but `method` is not.

This layer is applied *outside* `GovernorLayer` (`routes.rs:101` wraps the already-rate-limited `public`), so even 429'd/405'd requests still record their method. `curl -X AAAA1 …`, `-X AAAA2`, … in a loop mints a new counter+histogram series each time → unbounded RSS growth, and every `/metrics` scrape re-serializes the full set.

Suggest bounding to a known-method allowlist (all routes are GET anyway):
```rust
let method = match *req.method() {
    Method::GET => "GET",
    Method::HEAD => "HEAD",
    Method::OPTIONS => "OPTIONS",
    _ => "other",
};
```
then pass `method` (`&str`) straight into `with_label_values`.

settlement-distributions/bid-distribution/src/generators/tests.rs
```diff
@@ -1421,22 +1436,516 @@ fn test_activating_bid_charge_distributed_to_activating_stakers() {
...
+    }
+
+    let actual = serde_json::to_string_pretty(&settlements).unwrap();
```
kronael, line 1770: 🤖 claude: **Golden test only serializes — the runtime deserialize path is uncovered**

This asserts on `to_string_pretty(&settlements)` only. The path that actually runs in production — merkle-generator deserializing the flattened, internally-tagged `ClaimDetail` from JSON via `read_from_json_file::<SettlementCollection>` — has no unit test. It works today (verified), but a future change to the `ClaimDetail`/`SettlementClaim` serde attrs could silently break the merkle read with nothing failing here. One line closes it:
```rust
let _: Vec<Settlement> = serde_json::from_str(&expected).unwrap();
```

kronael (dismissed):
just some Claude nits... i'd say those can be just addressed in some other PR or change or ignored...

api/src/metrics.rs
ochaloup: Fixed: the `method` label is now bounded to an allowlist in `track_metrics` — `GET`/`HEAD`/`OPTIONS` are recorded verbatim, anything else lands in `other`, so arbitrary request methods can no longer mint new counter/histogram series. Included in the upcoming push (branch freshly rebased onto main).

settlement-distributions/bid-distribution/src/generators/tests.rs
ochaloup: Fixed: the golden test now deserializes the fixture back into `Vec<Settlement>` before the snapshot string comparison, so the merkle-generator runtime read path of the internally-tagged `ClaimDetail` is covered. Passes without re-blessing the snapshot. Included in the upcoming push.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -55,705 +58,571 @@ impl fmt::Display for ResultSettlementClaims {
...
+                group.stake_authority,
+                active_stake,
+                activating_stake,
+                claim_amount,
+                accounts.clone(),
+            ));
+            amount += claim_amount;
+        }
+    }
+    Ok((claims, amount))
```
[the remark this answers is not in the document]
ochaloup: Intentional — declining for this PR. The per-group flooring replicates the production engine behavior, and this refactor is gated on byte-identical settlement output vs production (scripts/regression-test-settlements.sh); redistributing rounding remainders would change claim amounts and fail that gate. The dust (at most groups−1 lamports per pool) stays unfunded — stakers are never overcharged — and callers already rely on the `amount <= pool` invariant. If lossless distribution is wanted, it should be its own PR with a re-blessed regression baseline.

packages/validator-bonds-sanity-check/src/dtoSettlements.ts
```diff
@@ -119,6 +119,17 @@ export class StakeAccountClaim {
   @Transform(({ value }) => (value ? new PublicKey(value) : value))
   readonly stake_authority!: PublicKey
 
+  @Expose()
+  @IsBigInt()
+  @Transform(({ value }) => (value ? BigInt(value) : 0n))
+  readonly claim_amount!: bigint
```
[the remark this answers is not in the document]
ochaloup: Intentional lenient parse — declining (this was already raised and resolved as intentional in an earlier review round). The Rust decoder is the enforcing side of the contract: `claim_amount` is a required field there, and every pipeline that acts on settlements (merkle-generator, init/fund/claim) parses with it, so a claim missing the amount can never enter the system. Within the sanity checker itself, the only computation on `claim_amount` is the cross-validation in checkMerkleTree, which hard-fails when the settlement-source totals do not equal the merkle-tree totals — and the merkle side is Rust-produced, so a coalesced 0 cannot silently pass; it just surfaces as a totals mismatch. Strictness here would only improve the error message, not safety.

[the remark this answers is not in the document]
ochaloup: Duplicate of the thread above — declining as intentional; see https://github.com/marinade-finance/validator-bonds/pull/434#discussion_r3602669000 for the reasoning (Rust decoder is the enforcing parser; the checkMerkleTree totals cross-validation fails loudly on any coalesced value, so leniency costs error-message precision, not safety).

api/src/bin/api.rs
```diff
@@ -69,119 +70,34 @@ async fn main() -> anyhow::Result<()> {
...
-
-    warp::serve(routes).run(([0, 0, 0, 0], params.port)).await;
-
-    Ok(())
-}
+    let public_addr = SocketAddr::from(([0, 0, 0, 0], params.port));
+    let internal_addr = SocketAddr::from(([0, 0, 0, 0], INTERNAL_PORT));
+
+    let public_listener = TcpListener::bind(public_addr).await?;
+    let internal_listener = TcpListener::bind(internal_addr).await?;
```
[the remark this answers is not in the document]
ochaloup: Fixed in 3a5a4d7 — startup now bails out with a clear error when `--port` equals the internal port 9000, before any listener is bound.

api/src/metrics.rs
```diff
@@ -0,0 +1,117 @@
...
+        .psql_client
+        .simple_query("SELECT 1")
+        .await
+    {
+        Ok(_) => StatusCode::OK,
+        Err(err) => {
+            log::warn!("readyz DB connectivity check failed: {err}");
+            StatusCode::SERVICE_UNAVAILABLE
+        }
+    }
```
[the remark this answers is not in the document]
ochaloup: Fixed in 3a5a4d7 — the readyz DB check is wrapped in a 2s tokio timeout; expiry logs a warning and returns 503, same as a query error.

api/src/metrics.rs
```diff
@@ -0,0 +1,117 @@
...
+        .await
+        .psql_client
+        .simple_query("SELECT 1")
+        .await
+    {
+        Ok(_) => StatusCode::OK,
+        Err(err) => {
+            log::warn!("readyz DB connectivity check failed: {err}");
+            StatusCode::SERVICE_UNAVAILABLE
+        }
```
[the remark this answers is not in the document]
ochaloup: Duplicate of the earlier readyz thread — fixed in 3a5a4d7 (2s timeout around the SELECT 1, 503 + warn on expiry).

settlement-distributions/bid-distribution/src/generators/tests.rs
```diff
@@ -1814,6 +2327,23 @@ fn sum_claims_for_authority(
...
+}
+
+fn find_claim_by_authority<'a>(
+    settlement: &'a Settlement,
+    stake_authority: &Pubkey,
+    withdraw_authority: &Pubkey,
+) -> Option<&'a settlement_common::settlement_collection::SettlementClaim> {
+    settlement.claims.iter().find(|c| {
+        c.stake_authority == *stake_authority && c.withdraw_authority == *withdraw_authority
+    })
```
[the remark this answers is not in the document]
ochaloup: Fixed in 3a5a4d7 — the helper is split into kind-aware `find_fee_claim_by_authority` (requires `ClaimDetail::FeeDeposit`) and `find_staker_claim_by_authority` (requires `StakerPayout`); all fee/staker assertion sites updated accordingly.

settlement-distributions/institutional-distribution/src/settlement_generator.rs
```diff
@@ -119,53 +98,71 @@ fn merge_payouts(
...
-                claim_amount: payout.payout_lamports,
-            });
+            let claim = match payout.kind {
+                PayoutKind::Staker => SettlementClaim::staker_payout(
+                    payout.withdrawer,
+                    payout.staker,
+                    payout.stake_amount,
+                    0,
+                    payout.payout_lamports,
+                    payout.stake_accounts,
```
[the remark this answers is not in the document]
ochaloup: Intentional, now documented and locked by a test (3a5a4d7). The claim's stake fields are the arithmetic basis `claim_amount` was computed from, not the staker's on-chain stake state — institutional payouts are computed from effective stake only (activating stake earns nothing that epoch), so the activating basis is always 0. Passing through the snapshot activating value would give the field two different meanings across pipelines and break report tooling that keys on `activating_stake != 0` to detect PriorityFee settlements. The invariant is documented on `ClaimDetail::StakerPayout` and asserted by `test_staker_payout_activating_stake_is_always_zero`; snapshot-state analytics belong in the settlement-level `details` block instead.

settlement-distributions/settlement-common/src/settlement_collection.rs
```diff
@@ -14,12 +14,91 @@ pub struct SettlementClaim {
...
+    StakerPayout {
+        active_stake: u64,
+        activating_stake: u64,
+        #[serde(with = "map_pubkey_string_conversion")]
+        stake_accounts: HashMap<Pubkey, u64>,
+    },
+    FeeDeposit,
+    // Zero-amount placeholder distinguishing Marinade- vs ValidatorBond-funded merkle
+    // roots when claim amounts would otherwise coincide. Not a real payout.
+    Marker,
```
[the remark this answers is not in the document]
ochaloup: Deferred — Markers are produced internally by this repo's generators (never ingested from third parties) and tampered JSON would still fail merkle-root verification on-chain, so decoder-side enforcement is a hardening follow-up rather than a blocker. A TODO in the code marks it (3a5a4d7).

api/src/metrics.rs
```diff
@@ -0,0 +1,117 @@
...
+    match String::from_utf8(buffer) {
+        Ok(body) => (StatusCode::OK, body),
+        Err(err) => {
+            log::error!("Metrics buffer was not valid UTF-8: {err}");
+            (
+                StatusCode::INTERNAL_SERVER_ERROR,
+                "invalid metrics encoding".to_owned(),
+            )
+        }
+    }
```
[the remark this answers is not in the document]
ochaloup: Fixed in 3a5a4d7 — the handler now returns `TextEncoder::format_type()` (`text/plain; version=0.0.4`) as the Content-Type; asserted in the http_behavior metrics test.

scripts/generate-discord-public-report.bash
```diff
@@ -2,6 +2,9 @@
 
 set -e
 
+# Deterministic decimal-point formatting regardless of the caller's locale
+export LC_NUMERIC=C
```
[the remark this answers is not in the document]
ochaloup: Fixed in 3a5a4d7 — `LC_ALL` is unset before exporting `LC_NUMERIC=C`.

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -82,6 +82,24 @@ impl VoteAccountRewards {
...
+            ("mev", self.realized_mev_commission_dec()),
+            ("block", self.realized_block_commission_dec()),
+        ]
+        .into_iter()
+        .filter_map(|(kind, c)| match c {
+            Some(c) if c < Decimal::ZERO => Some((kind, c)),
+            _ => None,
+        })
+        .collect()
+    }
```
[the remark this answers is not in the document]
ochaloup: Fixed in 3a5a4d7 — `negative_realized_commissions` became `realized_commission_anomalies`, which also flags reward types where stakers received rewards while the validator gross is zero (logged as its own warning); tests cover the zero-gross case.

settlement-distributions/institutional-distribution/src/settlement_generator.rs
```diff
@@ -119,53 +98,71 @@ fn merge_payouts(
...
     }
 
     settlements
         .into_values()
         .map(|mut settlement| {
             sort_claims_deterministically(&mut settlement.claims);
+            settlement.claims_count = settlement.claims.len();
             settlement
         })
         .collect()
```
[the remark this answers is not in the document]
ochaloup: Fixed in 3a5a4d7 — settlements are sorted by `vote_account` before returning (claims inside each settlement were already sorted); ordering is asserted in the generator tests.

packages/validator-bonds-sanity-check/src/losslessJson.ts
```diff
@@ -0,0 +1,143 @@
...
+ */
+function createLosslessParser(opts?: TokenizerOptions): LosslessParserHandle {
+  const tokenizer = new LosslessTokenizer(opts)
+  const tokenParser = new TokenParser({ paths: ['$'] })
+
+  tokenizer.onToken = tokenParser.write.bind(tokenParser)
+  tokenizer.onEnd = () => {
+    if (!tokenParser.isEnded) tokenParser.end()
+  }
+  tokenParser.onError = tokenizer.error.bind(tokenizer)
```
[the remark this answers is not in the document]
ochaloup: Not correct for the pinned version — `@streamparser/json` 0.0.22 does declare `error(err: Error): void` as a public method (dist/mjs/tokenizer.d.ts:34) and implements it (dist/cjs/tokenizer.js:701); it sets the ERROR state and routes to `onError`, which is exactly the semantic this wiring wants for parser errors. Empirically: every test in `__tests__/losslessJson.spec.ts` constructs this parser — if `tokenizer.error` were unavailable, the `.bind()` would throw a TypeError in all of them; the suite passes. Resolving as a false positive (web-search-sourced, not source-verified).

packages/validator-bonds-sanity-check/__tests__/losslessJson.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+    await expect(readLargeJsonFileLossless(filePath)).rejects.toThrow()
+  })
+
+  it('readLargeJsonFileLossless rejects malformed content mid-file', async () => {
+    const dir = mkdtempSync(join(tmpdir(), 'lossless-'))
+    const filePath = join(dir, 'malformed.json')
+    // malformed early, large valid-looking tail — the parse must fail either way
+    writeFileSync(filePath, `{"a": !!! ${'x'.repeat(1024 * 1024)}`)
+    await expect(readLargeJsonFileLossless(filePath)).rejects.toThrow()
+  })
```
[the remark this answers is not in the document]
ochaloup: Fixed in 3e0514b — the fixture now places the malformed token after a 512 KiB valid prefix (and before a valid-looking tail), per the suggestion.

api/tests/http_behavior.rs
```diff
@@ -0,0 +1,258 @@
...
+                "a 429 from the limiter must still carry CORS headers",
+            );
+            saw_429 = true;
+        }
+    }
+    assert!(
+        saw_429,
+        "exceeding the public burst (30) from one IP must yield a 429",
+    );
+}
```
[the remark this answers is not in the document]
ochaloup: Hardened in 4bffb90 — the test now fires 90 concurrent requests against the 30-token burst, so absorbing the 60 excess tokens would require more than 2 seconds of refill mid-burst (vs ~333 ms before). That makes a false pass require a pathologically stalled runner rather than ordinary CI jitter, without plumbing a test-only limiter through the production `with_public_rate_limit` signature. If the test ever flakes despite this margin, injecting a configurable limiter is the documented next step.

ochaloup:
@kronael I would like to ask for another review here. I received quite extensive feedback from both the AI reviewer and CodeRabbit, so it took me some additional time to finalize the code.

I would really like to push this through so I can continue with the follow-up work on bonds. I will be out next week, but once I’m back, I would be happy to merge this and start using the adjusted version of bonds. If you can check this PR once again when you have time, I would be grateful :pray:

kronael (dismissed):

kronael (approved):

PR #436, opened 2026-05-29
Set min_yield_premium_over_ssr_pmpe=3bp [trivial]

kronael:
We are still it seems underperforming by a few basis points, which makes me think we need to add a slack to compensate.

### DIFF
--- a/settlement-config.yaml
+++ b/settlement-config.yaml
@@ -3,7 +3,7 @@ fee_config:
   min_fee_bps: 200
   max_fee_bps: 800
   # PMPE premium added to ssi/ssr target. Range: -0.1..0.1 (~±2% APY).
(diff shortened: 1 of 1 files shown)


settlement-config.yaml
```diff
@@ -3,7 +3,7 @@ fee_config:
   min_fee_bps: 200
   max_fee_bps: 800
   # PMPE premium added to ssi/ssr target. Range: -0.1..0.1 (~±2% APY).
-  min_yield_premium_over_ssr_pmpe: 0
+  min_yield_premium_over_ssr_pmpe: 5
```
kronael:
```suggestion
  min_yield_premium_over_ssr_pmpe: 0.0003
```

kronael:
Probably not needed... we'll yet have to see.

PR #438, opened 2026-06-01
20260601 max fee adj [GEN-7676]

kronael:
When some validators underperform (low credits, downtime), their per-validator fee is
  already floored at min_fee_bps, so their staker yield lands below SSR. This drags the
  aggregate post-fee pmpe below ssr_pmpe even though individual fee caps are applied
  correctly. The fix is a binary search over max_fee_bps: starting at the configured ceiling
  and bisecting down until aggregate post-fee pmpe ≥ ssr_pmpe, returning the highest fee that
  satisfies the floor. Adds calculate_bid_settlement_totals to compute the aggregate
  stake/rewards/fees needed for bisect feedback.


settlement-config.yaml
```diff
@@ -1,6 +1,6 @@
 ---
 fee_config:
-  min_fee_bps: 200
+  min_fee_bps: 0
   max_fee_bps: 800
   # PMPE premium added to ssi/ssr target. Range: -0.1..0.1 (~±2% APY).
   min_yield_premium_over_ssr_pmpe: 0
```
ochaloup, line 6: I have a bit of an issue grasping what these `min/max` parameters mean.
What influences what.

For example, do we still need `min_fee_bps` now that the SSR calculation is fixed? Could not be simplified by removing it completely?

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -93,6 +155,68 @@ pub fn generate_bid_settlements(
...
+            &fc,
+            stake_authority_filter,
+            ssr_pmpe,
+        )?;
+        let totals = calculate_bid_settlement_totals(&settlements);
+        if totals.stake.is_zero() {
+            return Ok(settlements);
+        }
+        let post_fee = (totals.rewards - totals.fees) / totals.stake * Decimal::ONE_THOUSAND;
+        let next = if ssr_pmpe <= post_fee {
```
ochaloup: Based on Claude review, this does not seem fully correct

> Unsatisfiable SSR floor returns the *highest*-fee settlement instead of the lowest
> 
> When **no** `max_fee_bps` in the range can bring `post_fee_pmpe >= ssr_pmpe`
(e.g. a low-reward epoch where even `min_fee_bps` — now `0` after the config change —
leaves staker yield below the floor), the feasible branch never runs. the infeasible
branch's `get_or_insert` populates `best_settlements` on the **first** iteration only,
which is `current == max_cap` (the search starts at the ceiling). the bisection then
walks downward, computing and discarding lower-fee settlement sets, and finally returns
the max-fee result via `.expect(...)`. that is the worst possible outcome: maximum fees
extracted in exactly the scenario where stakers are already below the ssr floor — the
opposite of what "enforce global ssr_pmpe floor" intends. the loop already computes the
lowest-fee (`min_cap`) candidate but throws it away.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -93,6 +155,68 @@ pub fn generate_bid_settlements(
...
+            overshoot = current;
+            (current + undershoot) / 2
+        } else {
+            best_settlements.get_or_insert(settlements);
+            undershoot = current;
+            (current + overshoot) / 2
+        };
+        let next = next.clamp(min_cap, max_cap);
+        if next == current {
+            break;
```
ochaloup, line 193: Shouldn't the `info!` be here rather than only after the `break`?

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -85,6 +85,68 @@ pub struct PriorityFeeSettlementDetails {
...
+        .and_then(|v| v.as_u64())
+        .unwrap_or(0);
+    let dao = value
+        .get("dao_fee_claim")
+        .and_then(|v| v.as_u64())
+        .unwrap_or(0);
+    marinade + dao
+}
+
+pub fn calculate_bid_settlement_totals(settlements: &[Settlement]) -> BidSettlementTotals {
```
ochaloup, line 98: This seems as Claude points down. Can you verify?

> Activating-only validators' priority fees excluded from floor aggregate

ochaloup (changes requested):
I'm again a little bit afraid of adding more functionality to `bidding.rs`, which was originally meant to be much simpler and only manage data recalculated from the auction into snapshot data. I understand that the snapshot is only available here, but I'm concerned about bloating the code again.

Can we at least make a small improvement to the structure so it's easier to understand and maintain? I'm currently in the middle of some refactoring, so I can maybe think about it more deeply, but still...

The PR already created the seam — `generate_bid_settlements` (policy/search wrapper) vs `generate_bid_settlements_worker` (snapshot → claims). Promote the wrapper + calculate_bid_settlement_totals + effective_fee math into a dedicated
  module (fee_adjustment.rs / ssr_floor.rs). bidding.rs goes back to being the "snapshot → claims" data layer; the policy becomes unit-testable without a snapshot.

settlement-config.yaml
kronael: At this point maybe yes,  Until Michael comes that he does need it for us to make some revenue...  The change is wrong and flagged correctly, thank you.  As for us keeping it, I'd keep it.

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: It depends on what correct means.  It should keep the lowest one, I'll check.

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: If its above, we'll just see next=current.  What information does it provide?

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: They shouldn't be really.  Claude does seem to not work rather it seems...

kronael:
I agree.  Note, however, that the main problems I am mostly encountering, as also seen by your coments, is the way "telemetry" is extracted than the files being big or so... how are the events represented, etc.

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: Thank you for checking this.  I added them now.

settlement-distributions/bid-distribution/src/generators/bidding.rs
ochaloup: Ok, I see.

ochaloup (approved):

PR #440, opened 2026-06-03
Set min_fee_bps to 0 in settlement-config.yaml [trivial]

kronael:
Need to cut minimal fees as otherwise we are not able to meet SSR.

### DIFF
--- a/settlement-config.yaml
+++ b/settlement-config.yaml
@@ -1,6 +1,6 @@
 ---
 fee_config:
-  min_fee_bps: 200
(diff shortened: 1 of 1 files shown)


ochaloup (approved):
Approving, if it’s decided to be needed, that this can be merged easily.

kronael:
This won't be needed.  As the main reason it did not work is that the current code is not working in case there is a lot of psr settlements and penalties and also that the TVL was undervalues by the re-delegation amount.

PR #441, opened 2026-06-03
20260603 simulate fee [GEN-7676]

kronael:
BigQuery sums active + deactivating stake per authority when computing Marinade PMPE. In Solana, deactivating stake is still fully effective — active already includes it — so this
  double-counts re-delegation churn. simulate-fee was using active-only, producing a different denominator and a sign-flipped SSR comparison.

  To match BQ, bid-distribution now tracks total_marinade_redelegation_stake (deactivating stake under whitelist authorities, excluding real exits under exiting_stake_authorities) as a
  separate field in BidSettlementDetails. simulate-fee adds it to the TVL denominator. A new exiting_stake_authorities config in settlement-config.yaml marks authorities whose deactivating
  stake is real withdrawal, not churn.

  simulate-fee also gains -c mode (reads production settlement JSON from GCS, no CLI re-run) and prints marinade_stake_sol for direct cross-check against BQ.

We also add penalties to the calculation in the simulator.  Not in the optimization in bidding yet.


scripts/simulate-fee.ts
kronael, line 1: The duplicated types will be moved off in the followup PR.

scripts/simulate-fee.ts
```diff
@@ -114,139 +176,218 @@ const apy = (p: number, n: number) =>
...
+    const activeStake = bidDetails.reduce(
+      (sum, d) => sum + d.total_marinade_active_stake,
+      0,
+    )
+    const rustRedeleg = bidDetails.reduce(
+      (sum, d) => sum + (d.total_marinade_redelegation_stake ?? 0),
+      0,
+    )
+    const stakesPath = join(inp, 'stakes.json')
+    const redeleg =
```
ochaloup, line 345: I  think I commented this on #443 but noting here again. Feel free to merge and then talk about it in 443 or whatever.

Isn't there some duplication?

> 
**likely intent mismatch**: if the goal is "marinade tvl includes stake in transit that earns nothing during redelegation churn", the in-transit side at the epoch-e snapshot is the **activating** stake (re-staked on the new validator, not yet effective, earned nothing in e) — that is `total_marinade_activating_stake` in the rust worker (already computed, not exposed in details). the deactivating side earned rewards in e and is already in the denominator.

settlement-distributions/bid-distribution/src/settlement_config.rs
```diff
@@ -194,6 +202,10 @@ impl BidDistributionConfig {
         stake_authority_filter(self.whitelist_stake_authorities.clone())
     }
 
+    pub fn exiting_stake_authorities_filter(&self) -> Box<dyn Fn(&Pubkey) -> bool> {
```
ochaloup, line 205: Maybe discussed at #443 not sure what is the outcome

> Absent `exiting_stake_authorities` inverts semantics: everyone is treated as exiting

scripts/simulate-fee.ts
```diff
@@ -94,13 +124,45 @@ process.on('exit', () => {
...
+  whitelist_stake_authorities?: string[]
+  exiting_stake_authorities?: string[]
+  fee_config: { min_fee_bps: number; max_fee_bps: number }
+}
+
+function loadConfig(yaml: string): BidConfig {
+  return parseYaml(yaml) as BidConfig
+}
+
+async function redelegationStakeFromFile(
```
ochaloup, line 145: Claude notes

> File fallback triggers on a legitimate Rust 0 and computes a different metric
>
>  **Re-check**: `rustRedeleg > 0 ? rustRedeleg : ...` unchanged at lines 345-350.
> **Scope mismatch**: Rust sums deactivating stake only for validators that produced a Bidding settlement; `redelegationStakeFromFile` sums across the **entire** `stake_metas` array — including validators without settlements. The two values are not comparable; pick one scope and document it.

scripts/simulate-fee.ts
```diff
@@ -38,20 +58,30 @@ const { values, positionals } = parseArgs({
 })
 
 const [epochArg, ...feeStrs] = positionals
-let fees = feeStrs.map(Number)
+const fees: (number | null)[] = feeStrs.length ? feeStrs.map(Number) : [null]
```
ochaloup, line 61: nitpick from Claude

> Fee arguments are not validated; `NaN` flows into the YAML patch

scripts/simulate-fee.ts
```diff
@@ -114,139 +176,218 @@ const apy = (p: number, n: number) =>
...
-        '--data-dir',
-        dataDir,
-      ],
-      { stderr: 'pipe' },
+function runMkdir(dir: string) {
+  Bun.spawnSync(['mkdir', '-p', dir], { stderr: 'pipe' })
+}
+
+function runGcsCp(src: string, dst: string) {
+  Bun.spawnSync(['gcloud', 'storage', 'cp', src, dst], { stderr: 'pipe' })
```
ochaloup, line 187: Shouldn't this be more of a data preparation concern rather than part of the script? Something that gets prepared before the script is run?

ochaloup, line 187: Claude note:

> Non-zero exit code is not handled

ochaloup (approved):

scripts/simulate-fee.ts
kronael: Yes.  But this needs to be in-line with apy-api and other methodology, which uses active + deactivating.  The reason mostly being that we do not want to include the new stakers, which are activatig and delegated to the main staking authority, but deactivating on this authority are purely the re-delegation rotations since exiting stake is switched to the exit authority.

scripts/simulate-fee.ts
kronael: Solved in the #443.

scripts/simulate-fee.ts
kronael: This way, you can just come and run the simulation.  I won't see any benefit in splitting the script into more scripts.  But possibly I just missunderstand.

scripts/simulate-fee.ts
kronael: Ok.  Then the cli will fail.  That's correct and ok.  No need to complicate the code.

scripts/simulate-fee.ts
kronael: I added handling in #443

PR #442, opened 2026-06-03
[fix] derive onchain commissions for bid settlement claims from snapshot rewards [GEN-7037]

ochaloup:
The inflation/MEV commission claims used the auction-time onchain commission from sam-meta. A commission raised after the auction snapshot was applied to the epoch rewards but never charged to the bond, while `CommissionSamIncrease` stays blind to it for validators with in-bond commission set (effective commission = min(onchain, in-bond) does not change). The actual onchain commission is now derived from the snapshot rewards data as (gross - stakers) / gross, the same way the block rewards claim already works.

**Behavioral change:** rewards aggregation now hard-fails (previously `warn!` + skip) when any stake rewards entry cannot be mapped to a vote account, and when a `validators_*` rewards file has entries while its stakers' counterpart is empty. Skipped or missing stakers' rewards would inflate the derived `(gross - stakers) / gross` commission and overcharge bonds. Ops: watch the first production run for pipeline aborts from snapshot/rewards export skew; the fallback option is a materiality threshold.

https://github.com/marinade-finance/ds-sam/pull/119


ochaloup:
I rebased and fixed the `Cargo lint and test`

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -409,13 +467,21 @@ fn aggregate_rewards(
...
+            unmatched_jito = unmatched_jito.saturating_add(reward.amount);
             log::warn!(
                 "No vote account found for stake account {} in Jito priority fee rewards",
                 reward.stake_account
             );
         }
     }
 
+    // unmatched stakers' rewards would inflate the derived onchain commissions and overcharge validator bonds
+    if unmatched_inflation > 0 || unmatched_mev > 0 || unmatched_jito > 0 {
```
kronael, line 479: 🤖 claude: Hard-fail on unmatched stake accounts is a behavioral change not mentioned in the PR description — previously these were `warn!()` and skipped. The reasoning is sound (skipped stakers inflate `(gross - stakers) / gross`, overcharging bonds), but worth calling out explicitly so ops knows to watch the first production run for unexpected pipeline aborts from snapshot/rewards timestamp skew.

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -68,6 +69,31 @@ pub struct VoteAccountRewards {
...
+    // commission rate actually applied at rewards distribution; None when there were no rewards
+    pub fn realized_inflation_commission_dec(&self) -> Option<Decimal> {
+        realized_commission_dec(self.inflation_rewards, self.stakers_inflation_rewards)
+    }
+
+    pub fn realized_mev_commission_dec(&self) -> Option<Decimal> {
+        realized_commission_dec(self.mev_rewards, self.stakers_mev_rewards)
+    }
+
+    pub fn realized_block_commission_dec(&self) -> Option<Decimal> {
```
kronael, line 82: 🤖 claude: Uses `stakers_priority_fee_rewards` where the old block commission formula used `jito_priority_fee_rewards`. In `aggregate_rewards` both fields are always co-incremented so there's no behavioral difference now. If a default-constructed or partially-populated `VoteAccountRewards` ever goes through here the two can diverge silently. The fields are semantically identical — consider collapsing them or adding `debug_assert_eq!(self.stakers_priority_fee_rewards, self.jito_priority_fee_rewards)`.

settlement-distributions/bid-distribution/src/generators/tests.rs
```diff
@@ -830,6 +836,163 @@ fn test_zero_rewards() {
...
+    // distinct per-type rates kept onchain above the 0% in-bond promise
+    assert_eq!(
+        inflation_claim,
+        Decimal::from(inflation_rewards) * Decimal::from_str("0.05").unwrap(),
+        "Inflation claim must cover the commission raised after the auction snapshot"
+    );
+    assert_eq!(
+        mev_claim,
+        Decimal::from(mev_rewards) * Decimal::from_str("0.08").unwrap(),
+        "MEV claim must cover the commission raised after the auction snapshot"
```
kronael, line 934: 🤖 claude: Block commission claim is not asserted in this test — since no block rewards are set in the `RewardsParams`, `realized_block_commission_dec()` returns `None` → `Decimal::ZERO`. An explicit `assert_eq!(block_claim, Decimal::ZERO)` would document that the zero result is intentional rather than silently unchecked.

kronael:
🤖 claude: Minor ops note: `CLAUDE.md` at the repo root describes `rewards.rs` as _"parses inflation/MEV/block rewards input files"_. After this PR it also derives realized commission rates from those files (`(gross - stakers) / gross`), so the description is slightly incomplete. Might be worth a one-line update.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -712,3 +658,17 @@ fn generate_bid_settlements_worker(
...
+
+// charges the realized commission from snapshot rewards; the auction-time sam-meta value misses commission raised after the auction snapshot
+fn commission_claim(
+    commission_realized_dec: Option<Decimal>,
+    commission_in_bond_dec: Decimal,
+    marinade_rewards: Decimal,
+) -> Decimal {
+    match commission_realized_dec {
+        Some(realized) if realized > commission_in_bond_dec => {
+            marinade_rewards.mul(realized - commission_in_bond_dec)
```
kronael: The multiplication does not really need to be in the function.  It would be enough to have a "commision_eff" and then X.mul(Comm) ... Comm is 0 if anything is wrong so X.mul(0) = 0.

settlement-distributions/bid-distribution/src/generators/tests.rs
```diff
@@ -837,6 +836,163 @@ fn test_zero_rewards() {
     );
 }
 
+#[test]
+fn test_commission_raised_after_auction_charged_from_rewards() {
+    // auction snapshot saw 0% onchain commission but rewards were distributed at 5%/8%;
```
kronael, line 855: Does this whole thing now replace the "ds-sam evaluate-revenue-changes"?  I like it more here, for sure.

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -68,6 +69,31 @@ pub struct VoteAccountRewards {
...
+        realized_commission_dec(self.block_rewards, self.stakers_priority_fee_rewards)
+    }
+}
+
+fn realized_commission_dec(gross_rewards: u64, stakers_rewards: u64) -> Option<Decimal> {
+    if gross_rewards == 0 {
+        return None;
```
kronael, line 89: Here it might be correct to have 1 also, which would simplify the function...

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -281,6 +314,26 @@ pub fn load_rewards_from_directory(
     })
 }
 
+// an empty stakers' rewards file with a populated validators' counterpart would derive 100% commissions and overcharge bonds
```
kronael, line 317: How could this happen?  If the file is non-empty and json - then how do we say its not correct?  Would it perhaps make sense to err-out on any empty file?

kronael (dismissed):
There are some "maybe simplify this" comments, but mostly nits.  Claude added some, please feel free to comment on the usefulness.

settlement-distributions/bid-distribution/src/rewards.rs
ochaloup: I know this was a `warn!` before, but I believe we should make it a hard failure. I hope it works correctly and helps enforce consistency.

settlement-distributions/bid-distribution/src/rewards.rs
ochaloup: I think `None` is more appropriate here. I'm leaving it as is, but feel free to reopen if you consider it important.

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -281,6 +314,26 @@ pub fn load_rewards_from_directory(
...
+fn verify_stakers_rewards_present(
+    inflation_rewards: &[StakeRewardEntry],
+    mev_rewards: &[StakeRewardEntry],
+    validators_inflation: &[VoteRewardEntry],
+    validators_mev: &[VoteRewardEntry],
+) -> anyhow::Result<()> {
+    if !validators_inflation.is_empty() && inflation_rewards.is_empty() {
+        return Err(anyhow::anyhow!(
+            "{VALIDATORS_INFLATION_REWARDS_FILE} has entries but {INFLATION_REWARDS_FILE} is empty - stakers' inflation rewards export is incomplete"
+        ));
```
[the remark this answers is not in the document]
ochaloup: Reworded to cover both causes: "… is empty - either the stakers' inflation rewards export is incomplete or no stakers received inflation rewards; refusing to derive 100% commissions".

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -281,6 +314,26 @@ pub fn load_rewards_from_directory(
...
+    }
+    if !validators_mev.is_empty() && mev_rewards.is_empty() {
+        return Err(anyhow::anyhow!(
+            "{VALIDATORS_MEV_REWARDS_FILE} has entries but {MEV_REWARDS_FILE} is empty - stakers' MEV rewards export is incomplete"
+        ));
```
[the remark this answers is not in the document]
ochaloup: Reworded the same way for MEV: "… either the stakers' MEV rewards export is incomplete or no stakers received MEV rewards; refusing to derive 100% commissions".

settlement-distributions/bid-distribution/src/generators/tests.rs
ochaloup: Added an explicit assert_eq!(block_claim, Decimal::ZERO, ...) at the end of the test to document that the zero block claim is intentional (no block rewards configured in this test).

settlement-distributions/bid-distribution/src/generators/bidding.rs
ochaloup: Refactored as suggested: commission_eff(realized, in_bond) now returns just the effective rate and the three call sites do marinade_*_rewards.mul(commission_eff(...)).

ochaloup:
Re the CLAUDE.md suggestion (https://github.com/marinade-finance/validator-bonds/pull/442#issuecomment-4622263049): there is no CLAUDE.md tracked in this repository and the quoted rewards.rs description does not appear in any tracked file — likely a local untracked file, so nothing to update here.

settlement-distributions/bid-distribution/src/rewards.rs
ochaloup: Distincts concepts

settlement-distributions/bid-distribution/src/generators/tests.rs
ochaloup: No, they're complementary layers. ds-sam analyze-revenues still produces revenue-expectation.json, which feeds the PSR/protected-events path

settlement-distributions/bid-distribution/src/rewards.rs
ochaloup: This is kind of a "smoke test failure" to check for the most obvious errors - the stakers' and validators' exports are separate files from separate steps, so a partial/aborted export can leave a structurally valid but empty [] stakers file whose counterpart has entries.

ochaloup:
@kronael Thank you for the review! I updated the branch, can you please re-check?

kronael (dismissed):
Thank you for the good work.

ochaloup:
@kronael merge conflict resolved to fit to the changes in `main/HEAD`. Can you re-approve please?

settlement-distributions/bid-distribution/src/rewards.rs
```diff
@@ -409,13 +467,21 @@ fn aggregate_rewards(
...
             );
         }
     }
 
+    // unmatched stakers' rewards would inflate the derived onchain commissions and overcharge validator bonds
+    if unmatched_inflation > 0 || unmatched_mev > 0 || unmatched_jito > 0 {
+        return Err(anyhow::anyhow!(
+            "Unmatched stake accounts in rewards files (lamports): inflation {unmatched_inflation}, mev {unmatched_mev}, jito priority fee {unmatched_jito}"
+        ));
+    }
```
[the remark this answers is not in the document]
ochaloup: zero-lamport entries are inert

kronael (approved):

PR #443, opened 2026-06-03
20260603 annotate fee [trivial]

kronael:
Make the appropriate correctness changes also in the annotate fee script.


scripts/fee-annotation.ts
```diff
@@ -56,7 +76,10 @@ async function main() {
     (s): s is BidSettlement => s.reason === 'Bidding' && s.details !== null,
   )
   const stake = bids.reduce(
-    (s, b) => s + b.details.total_marinade_active_stake,
+    (s, b) =>
+      s +
+      b.details.total_marinade_active_stake +
+      (b.details.total_marinade_redelegation_stake ?? 0),
```
ochaloup, line 56: Don't we double-count the deactivated lamports?

scripts/simulate-fee.ts
ochaloup: Claude points

> Fallback trigger `rustRedeleg > 0` misfires on legitimate zero; fallback overcounts
>
> Two problems:
> 1. **Wrong trigger.** The fallback exists for old production files where the field is absent, but it fires whenever the summed value is `0` — including when the field is present and legitimately zero (e.g. all deactivating stake under exiting authorities, or simply no churn). Detect absence instead: `bidDetails.some(d => d.total_marinade_redelegation_stake === undefined)`.
> 2. **Different semantics.** `redelegationStakeFromFile()` sums deactivating lamports across **all** stake metas in `stakes.json` — including validators that produced no Bidding settlement (skipped zero-marinade-stake validators, non-SAM validators) — while the Rust field only covers validators present in `bidSettlements`. In `-c` mode it additionally applies the *current repo's* whitelist/exiting config to historical epochs. The fallback systematically overcounts relative to the field it substitutes for. At minimum, restrict the file scan to vote accounts present in `bidSettlements` (the file's stake metas carry `validator`), and emit a `# redeleg from stakes.json (approx)` stderr note so output is not mistaken for exact.
>

ochaloup, line 40: Claude mentions

> `-c` mode evaluates historical prod settlements against the current repo's fee config
> 
> In `-c` mode `maxFee`/`minFee` come from the working tree's `settlement-config.yaml`, but the downloaded `bid-distribution-settlements.json` was generated with whatever fee config production used at that epoch. `pmpeMax`, `fee_sol_max`, `validators_capped`, and `validators_at_min_fee` are then computed/classified against potentially wrong bps — for old epochs spanning fee changes the "capped" counts are meaningless. Relatedly, the guard rejects positional fee args in `-c` mode but silently honors `-m`, which re-patches `min_fee_bps` and shifts the `nAtMin` classification.

scripts/fee-annotation.ts
```diff
@@ -136,9 +159,26 @@ async function main() {
...
   const epochsPerYear = SECONDS_PER_YEAR / (cur.time - prev.time)
 
+  const psrToStakers = settlements.reduce(
+    (sum, s) => (isProtectedEvent(s.reason) ? sum + s.claims_amount : sum),
+    0,
+  )
+  const penaltyToStakers = settlements.reduce((sum, s) => {
+    const d = s.details
+    if (s.reason === 'BidTooLowPenalty')
+      return sum + (d?.stakers_bid_too_low_penalty_claim ?? 0)
```
ochaloup: Claude points

> Penalty staker extras use cap fields, not actual claimed amounts
>
> `stakers_bid_too_low_penalty_claim` / `stakers_blacklist_penalty_claim` / `stakers_bond_risk_fee_claim` are upper bounds — `sam_penalties.rs:189-209` only `ensure!`s `claimed_* <= stakers_*`. When some stake accounts are not claimable, the actual distribution is lower, so `penalty_sol_to_stakers` and the pmpe boost are slightly overestimated. The ProtectedEvent path correctly uses `claims_amount` (actual). If precision matters, sum per-settlement actual staker claims instead; otherwise add a comment noting these are caps.

scripts/simulate-fee.ts
ochaloup: Claude points

> BidDetails.total_marinade_redelegation_stake` typed required but absent in old prod files
> 
> `-c` mode reads production settlements generated before the field existed, and line 344 already defends with `?? 0` — but the type says `number`, making the `??` look dead and the type a lie. Change to `total_marinade_redelegation_stake?: number` (as `fee-annotation.ts:22` already does). This also unblocks the field-absence detection proposed in P2-R1#4.

ochaloup (commented):
Mostly worked with Claude to determine some questions around the code

There seems to be some code duplication around?

> The `Reason` union, `isProtectedEvent`, the penalty-extras reducer (identical 10-line `reduce`), and the `feesByVote`/`feesByValidator` per-vote-account fee aggregation are copy-pasted across both scripts (fee-annotation.ts:7-36,106-123,162-176 vs simulate-fee.ts:12-46,368-396). Extract into e.g. `scripts/settlement-utils.ts` (`sumStakerExtras(settlements)`, `feesByVoteAccount(settlements)`, shared `Settlement`/`Reason` types). Note: no `@marinade.finance` typescript-common runtime package is currently in `package.json` (only `eslint-config` and `web3js-1x-testing`) — worth considering whether these settlement-JSON helpers belong in typescript-common if other repos parse the same files.

kronael:
Well those are utility script.  I don't see much use in building some utilities around if there is no packaging etc for typescript.  Or you mean like really creating scripts/utils.ts?  That seems off.  If so, I'd need to rather create some full fledged ts package or project and work in there.  Or am I just not getting it?

scripts/fee-annotation.ts
kronael: Oh god.  I don't really get it much though.  I am also using mostly Claude.  So... he should get it.

scripts/simulate-fee.ts
kronael: Yes.  This is a feature.  Would be best if the settlement included the fee, but this is not the case.

scripts/fee-annotation.ts
kronael: I would hope not as this was the whole point. But working with Claude lately has become really frustrating.  Its painstakingly slow and mostly just wrong every time.

kronael: 🤖 claude: Not double-counting — `total_marinade_redelegation_stake` is added once per bid settlement, same as `total_marinade_active_stake`. Both are included in the stake denominator intentionally: redelegating (deactivating) stake still earns rewards and contributes to Marinade's fee base. Matches the fix in `bidding.rs` where the feasibility denominator was updated to include redelegation for the same reason.

scripts/simulate-fee.ts
kronael: 🤖 claude: Both points valid. Will fix: (1) detect field absence (`=== undefined`) rather than zero value; (2) emit a stderr note marking the fallback value as approximate. Done in 49597b3.

kronael: 🤖 claude: Intentional — the purpose of simulate-fee is to evaluate what a proposed config *would have done* on historical data, so using the current repo config against prod settlements is the desired behavior. The `-m` flag silently applying in `-c` mode is a fair catch though — guarded in 49597b3 to reject it the same way positional fee args are rejected.

scripts/fee-annotation.ts
kronael: 🤖 claude: Valid. Already switched simulate-fee.ts to use `claims_amount` for all penalty types (bfb17b6). Same fix applied to fee-annotation.ts in 49597b3.

scripts/simulate-fee.ts
kronael: 🤖 claude: Valid, fixed in 49597b3 — `total_marinade_redelegation_stake?: number`. The runtime `?? 0` already handled the absent case correctly; the type just needed to match.

kronael:
Ok.  Once ok, I'd move to the other branch #441 .

kronael:
I am also adding maybe some of the utilities to typescipt-common...  Would need to have the #444 ideally in tomorrow, if we can mange it, so lets see.  I am sorry but I ignored too many things and hoped it would deliver a good-enough solution, but it did not, so need to push the other changes too.

scripts/fee-annotation.ts
ochaloup: The question is rather: does `total_marinade_active_stake` contain the deactivating stake?

First, let's verify whether it does (I believed it does, but I may be wrong). If so, then we should consider whether there is a double calculation.

kronael: It does include it.  But still we need to add it so as to amortize the re-delegating stake.  Otherwise the APY is wrong.  This is because we do not include inactive stake in TVL at the epoch when it is inactive, but using this addition we include it in the epoch during which it is being deactivated.  This makes it in-total-the-same.

scripts/settlement-utils.ts
```diff
@@ -0,0 +1,91 @@
+// Shared settlement-JSON types and helpers for off-chain distribution scripts.
+// TODO: move to @marinade.finance/ds-sam once that package exists.
```
kronael, line 2: Where should we move this?  Claude says not to validator-bonds.  But maybe we should have it elsewhere than ds-sam.

kronael:
Now truly ready for re-review please.

scripts/simulate-fee.ts
ochaloup: Ok

scripts/settlement-utils.ts
ochaloup: I think the settlement distribution should live in validator-bonds. Maybe we already have some kind of "kickoff" structure here:
https://github.com/marinade-finance/validator-bonds/blob/main/packages/validator-bonds-sanity-check/src/dtoSettlements.ts

That's not the right place for something generic and reusable, though. I would move it under `packages/` as something like `settlement-common`, `pipelines-common`, or similar.

scripts/settlement-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+      reason: ProtectedEventReason
+      vote_account: string
+      claims_amount: number
+      details: null
+    }
+
+export type BidSettlement = Extract<Settlement, { reason: 'Bidding' }>
+
+/** Sum of PSR + penalty lamports redistributed to stakers (uses claims_amount). */
+export function sumStakerExtras(settlements: Settlement[]): number {
```
ochaloup, line 70: Claude notes

> `sumStakerExtras` sums `claims_amount` for the three penalty reasons, but a **BidTooLowPenalty** settlement's `claims_amount` includes the marinade and DAO fee claims pushed into the same claims vec (sam_penalties.rs:217-238: `claimed_bid_too_low_penalty_amount += marinade_...` and `+= dao_...`). So `stakerExtras` → `pmpeAdj`, `pmpeMax`, `penalty_sol_to_stakers` are overstated by the marinade+DAO portions of bid-too-low penalties. Verified the other reasons are clean: BlacklistPenalty and BondRiskFee accumulate staker claims only (sam_penalties.rs:174,185); ProtectedEvent settlements are staker-only with `details: None` (psr_events.rs:69-127). The R2 claim that the claims_amount switch was a clean fix was wrong for this one reason.

scripts/simulate-fee.ts
```diff
@@ -311,28 +292,30 @@ for (let epoch = epochStart; epoch <= epochEnd; epoch++) {
...
+    const bidSettlements = settlements.filter(
+      (s): s is BidSettlement => s.reason === 'Bidding' && s.details !== null,
+    )
+    const bidDetails = bidSettlements.map(s => s.details)
     if (!bidDetails.length) {
       process.stderr.write(`  # no data for fee=${fee} epoch=${epoch}\n`)
       continue
     }
 
+    // Deactivating (redelegating) stake is included in the TVL denominator
```
ochaloup: Isn't this wrongly stated?

scripts/settlement-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+  total_marinade_active_stake: number
+  total_marinade_redelegation_stake?: number
+  total_marinade_stakers_rewards: string
+}
+
+type PriorityFeeDetails = FeeDetails & {
+  total_marinade_active_stake: number
+}
+
+type PenaltyDetails = Record<string, never>
```
ochaloup: Is `never` really a good type here?

ochaloup (changes requested):
Please fix the CI test failures.

Otherwise, Claude has objections to the calculation of `sumStakerExtras`. Other than that, just two nitpicks.

scripts/settlement-utils.ts
kronael: It does, its a bug, we either will remove it or ignore it.  Its not part of the bisection and its ok like it is as it is.

scripts/simulate-fee.ts
kronael: Most comments are useless.  Claude just keeps adding them.  Sorry.

kronael:
Thank you for the review.  I am sorry but there is something yet wrong.  Fixing it now.

kronael:
@ochaloup If I could ask for another review now.

kronael:
Ok.  Hope this now fixes the Lint.  As I was not aware we changed TS, but apparently we did...

kronael:
Now passes.

ochaloup (approved):

PR #444, opened 2026-06-03
20260603 improve opt [GEN-7676]

kronael:
Fee bisection was computing post-fee yield from bid rewards and fees alone, ignoring the lamports that penalty and PSR settlements redistribute to stakers. This understated actual yield,
  causing the bisector to set lower fees than the SSR floor allowed — leaving staker budget on the table.
 
  The bisection now feeds total_staker_extras (penalty + PSR staker claims, computed before the bid step runs) into the feasibility check, so post_fee_pmpe = (rewards + staker_extras − fees)
  / stake correctly reflects what stakers receive. A two-phase strategy runs max-fee bisection first; if max_fee pins at the configured ceiling with post-fee still above the SSR target, a
  second phase raises min_fee to extract the remaining headroom. If the target can never be met, min-cap settlements are returned as fallback.
  
  Redelegation stake and deactivating stake are also added to the TVL denominator used in PMPE calculations, fixing an undercount that inflated per-unit yield estimates. The adjusted fee
  bounds (adj_max_fee_bps, adj_min_fee_bps) and ssr_pmpe are now surfaced in SettlementCollection, letting the simulation scripts derive capped/atMin diagnostics directly from the settlement
  JSON rather than re-running the CLI.


settlement-distributions/bid-distribution/src/generators/sam_penalties.rs
```diff
@@ -42,6 +42,24 @@ pub struct BondRiskFeeDetails {
     pub stakers_bond_risk_fee_claim: u64,
 }
 
+/// Total lamports the penalty settlements redistribute to stakers (excludes the
+/// Marinade/DAO distributor cut). Fed to the bid bisection so post-fee yield
+/// reflects penalty payouts.
+pub fn calculate_total_penalties(settlements: &[Settlement]) -> Decimal {
```
kronael, line 48: Again bid too low is correctly here.

settlement-distributions/bid-distribution/src/bin/cli.rs
```diff
@@ -109,6 +113,64 @@ fn main() -> anyhow::Result<()> {
     let mut adj_min_fee_bps: Option<u64> = None;
     let mut collection_ssr_pmpe: Option<f64> = None;
 
+    // ===== PSR Settlements (Protected Events) =====
```
kronael, line 116: 🤖 claude: PSR block moved here from after the SAM block. It previously ran last; it now runs first so `total_staker_psr_settlements` is available when `generate_bid_settlements` is called — the bisection needs it to compute full post-fee staker yield.

settlement-distributions/bid-distribution/src/bin/cli.rs
```diff
@@ -217,68 +261,32 @@ fn main() -> anyhow::Result<()> {
             "Generated {} penalty settlements",
             penalty_settlements.len()
         );
+        let total_staker_penalties = calculate_total_penalties(&penalty_settlements);
```
kronael, line 264: 🤖 claude: Penalty settlements moved before the bid call (was after). `total_staker_penalties` must be ready to pass into `generate_bid_settlements` as part of `total_staker_extras`.

settlement-distributions/bid-distribution/src/bin/cli.rs
```diff
@@ -217,68 +261,32 @@ fn main() -> anyhow::Result<()> {
...
-            &stake_authority_filter,
-            &psr_configs,
+            &sam_validator_metas,
+            &rewards_collection,
+            bidding_config,
+            &bid_distribution_config.fee_config,
+            &*stake_authority_filter,
+            &*exiting_stake_authority_filter,
+            ssr_pmpe,
+            total_staker_penalties + total_staker_psr_settlements,
```
kronael, line 278: 🤖 claude: All staker-redistributed lamports (penalty payouts + PSR payouts) passed to the bisection here. The post-fee PMPE formula in `bidding.rs` adds these to bid rewards before checking the SSR floor, so the bisection sees what stakers actually receive — not just bid yield.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -156,18 +160,28 @@ pub fn generate_bid_settlements(
     stake_authority_filter: &dyn Fn(&Pubkey) -> bool,
     exiting_stake_authority_filter: &dyn Fn(&Pubkey) -> bool,
     ssr_pmpe: Decimal,
+    total_staker_extras: Decimal,
```
kronael, line 163: 🤖 claude: New param: `calculate_total_penalties(&penalty_settlements) + calculate_total_psr_staker_claims(&psr_settlements)` from the call site. Used in the feasibility check as `(rewards + total_staker_extras − fees) / stake` so the SSR floor comparison accounts for full staker yield, not just bid rewards.

kronael:
@ochaloup please, this would seem is now ready for review.  A lot of the changes are blocks of code moved around which I tried to annotate with claude so its intelligible.

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -182,37 +196,45 @@ pub fn generate_bid_settlements(
...
-        current = next;
+        break;
     }
-    let settlements = best_feasible.or(best_infeasible).expect("MAX_ADJ_ITER = 0");
     Ok(BidSettlementValues {
-        settlements,
-        adj_max_fee_bps: best_feasible_fee,
-        adj_min_fee_bps: min_cap,
+        settlements: best.or(fallback).expect("MAX_ADJ_ITER = 0"),
+        adj_max_fee_bps: overshoot,
```
ochaloup: claude

> When phase 1 converges feasible at max_cap and phase 2 starts raising `min_fee_bps`, overshoot is reset to track the min axis — but it's still returned as `adj_max_fee_bps`. Production JSON metadata reports a `min-fee` value labeled as max in exactly the feature's primary scenario. Downstream: simulate-fee.ts -c validators_capped counts go wrong. No test covers the phase-2 path or asserts `adj_*` outputs.
> Fix: `capture adj_max = current` before resetting the bracket; use overshoot (not fc.min_fee_bps) for `adj_min_fee_bps`; add a phase-2 test.

settlement-distributions/bid-distribution/src/generators/sam_penalties.rs
ochaloup: Claude

> calculate_total_penalties sums full claims_amount for BidTooLowPenalty, which includes Marinade/DAO distributor claims — then it's added to the staker post-fee PMPE numerator vs the SSR target. Overstates staker yield → bisection keeps higher fees than the SSR-floor promise permits. May be deliberate ("redistributed within the ecosystem"), but conflicts with the PSR-claims rationale. Either document the economics or sum only stakers_* fields; keep the TS mirror sumStakerExtras and its "to stakers" labels in sync.

ochaloup (dismissed):
Seems fine to me in general. Just two points from claude.

settlement-distributions/bid-distribution/src/generators/bidding.rs
kronael: This seems to be a big bug.  I will  need to make more sure I run one more round of claude over it or I can't see why he does not seem to catch these obvious things.

kronael:
@ochaloup ... fixed the bug you found - thank you.  Now ready for the blessing.

ochaloup (approved):

PR #446, opened 2026-06-05
20260604 analytics [GEN-7676]

kronael:
:robot: 
  The production settlement JSON was reporting adj_max_fee_bps and adj_min_fee_bps incorrectly
  when the phase-2 min_fee bisect ran: overshoot tracks the active axis, so after phase 2
  starts it tracks the min_fee axis rather than the max_fee axis, causing the metadata to
  label a min-fee value as the max. The same bisect also stored the last probed (potentially
  infeasible) fc.min_fee_bps as adj_min rather than the highest feasible value.

  Fixed by capturing adj_max = current before resetting the bisect bracket on phase-2 entry,
  and returning overshoot as adj_min_fee_bps at convergence. Also adds log lines at the phase
  switch and at the final result so the bisect is traceable in simulation output.

  simulate-fee.ts previously fetched inputs by running regression-test-settlements.sh, which
  downloads ~5 GB per epoch (input files plus expected/actual regression artifacts already
  synced to GCS). It now fetches the 10 input files directly from GCS and the scoring API,
  stores them gzip-compressed, and decompresses to the OS temp directory (tmpfs) before each
  CLI run. This cuts per-epoch disk use from 5 GB to ~400 MB. Shell helpers are renamed to the
  run prefix and hard-exit on non-zero codes rather than returning booleans.

  simulate-fee.ts also gains a pmpeAdj fix: activating-only validators (no Bidding settlement,
  only PriorityFee) had their activating_bid_claim missing from totalRewards while their fees
  were still subtracted, deflating the computed staker yield.
 
  report-chart.ts is a new TypeScript chart script (replacing the Python matplotlib version)
  built on vega-lite + sharp. It produces a five-panel PNG from report.yml:

  - Post-fee APY vs SSR baseline with green/red fill showing per-epoch overshoot/undershoot,
  ±pp delta labels at extremes, and gap markers where simulations were skipped
  - Monthly (or weekly for short ranges) fee extraction bars with a 3000 SOL/month target
  line, paired with a monthly shortfall panel; incomplete periods render as hollow dashed bars
  - Validators at cap / at min fee over epochs
  - Optimized fee floor and ceiling (adj_min/adj_max) per epoch
  - An isolated legend row between the APY panel and the bar panels


settlement-config.yaml
kronael, line 1: These will be subject to tuning before we're finished.

kronael:
@ochaloup this one will make it possible for Marinade to earn 200SOL each epoch and tune the fee accordingly.

scripts/simulate-fee.ts
```diff
@@ -114,11 +161,17 @@ function tmpFile() {
...
 
 type BidConfig = {
   whitelist_stake_authorities?: string[]
   exiting_stake_authorities?: string[]
-  fee_config: { min_fee_bps: number; max_fee_bps: number }
+  fee_config: {
+    min_fee_bps: number
+    max_fee_bps: number
+    min_yield_premium_over_ssr_pmpe?: number
+    target_sol_revenue?: number
```
ochaloup: Claude

> simulate-fee.ts writes target_sol_revenue: into YAML but Rust field is min_sol_revenue; no deny_unknown_fields → -s/--target-sol silently ignored

settlement-distributions/bid-distribution/src/generators/bidding.rs
```diff
@@ -212,36 +239,62 @@ pub fn generate_bid_settlements(
...
+    } else {
+        adj_phase1
+    };
+    let adj_min_fee_bps = if !tuning_max {
+        overshoot
+    } else if min_first {
+        adj_phase1
+    } else {
+        fc.min_fee_bps
+    };
```
ochaloup, line 292: Isn't this a bit long code for if/else/case?

scripts/simulate-fee.ts
```diff
@@ -323,38 +486,44 @@ for (let epoch = epochStart; epoch <= epochEnd; epoch++) {
...
-    const redeleg = !fieldAbsent
-      ? rustRedeleg
-      : existsSync(stakesPath)
-        ? await redelegationStakeFromFile(stakesPath, cfg)
-        : 0
-    if (fieldAbsent && redeleg > 0)
-      process.stderr.write('  # redeleg from stakes.json (approx)\n')
+    const stakesPath = join(inp, STAKES_FILE)
+    const redeleg =
+      rustRedeleg > 0
```
ochaloup, line 528: Claude

> redeleg fallback rustRedeleg > 0 ? … : fileApprox overrides a legitimate zero
>  Gate on field presence, not > 0

scripts/report-chart.ts
```diff
@@ -0,0 +1,746 @@
...
+  return parseFloat(String(s).replace('%', ''))
+}
+
+function ratio(s: string | number | undefined): number {
+  if (!s) return 0
+  const [a, b] = String(s).split('/').map(Number)
+  return b ? (a / b) * 100 : 0
+}
+
+function isoWeek(ts: number): string {
```
ochaloup, line 88: Claude

> isoWeek (lines 88-95) uses local-timezone getters — d.getFullYear(), jan4.getDay(), jan4.getDate(), and d.getTime() anchored to a local-constructed new Date(year, 0, 4). isoMonth (lines 97-99) uses UTC via toISOString()

scripts/report-chart.ts
```diff
@@ -0,0 +1,746 @@
...
+                  strokeWidth: 1.5,
+                  strokeDash: [4, 3],
+                },
+                encoding: { y: { datum: 3000 } },
+              },
+              {
+                data: { values: [{}] },
+                mark: {
+                  type: 'text',
+                  text: '3000 SOL target',
```
ochaloup: Not sure here, but it seems like a hard-coded value here for the chart?

ochaloup (dismissed):
It seems clean and alright to me; nice!

I was just wondering a bit about the change in direction from PNPME to more for the SOL calculations. But seems all reasonable. And I’m personally not able to evaluate the details, so I discussed this with Claude. His hints, just to note them here:

> The bits I find subtle / worth knowing
> 
> 1. Both modes ultimately speak target_pmpe. The "SOL revenue" mode isn't a different worker — it's the same per-validator math, fed a target_pmpe derived backwards from the SOL goal. That's elegant but means the SOL→PMPE conversion (and its clamp(0,1)) is load-bearing; get that wrong and every fee is wrong.
> This is exactly where the SOL-mode bugs in your commit history were (clamp ratio not pmpe, include extras).
> 2. The probe exploits fee-independence (my P3-R1#5). The conversion needs settlement_sol and total_stake, which don't depend on fees — so cli.rs runs a throwaway target_pmpe=0 pass just to read them. Correct, but it's running the whole bisection for two invariant numbers.
> 3. Fallback semantics. If the target can never be met (e.g. SOL target exceeds all available rewards), best stays None and it returns the last infeasible settlements (fallback) — i.e. max-extraction best-effort. The reported adj_*_fee_bps in that corner can be the initial bound rather than the fallback's
> actual config, which is a diagnostic-only imprecision, not a payout bug.
> 4. best tracks overshoot for free because they're assigned together in the if feasible branch — so the returned settlements always correspond to the reported feasible bound, even across the phase-1→phase-2 reset (phase 2's first probe re-establishes feasibility immediately).

A tech note: can you please run `cargo clippy` and fix the issues? It seems to be failing.

PR #360, opened 2026-03-12
[docs] add CLAUDE.md, skill, fix READMEs [GEN-7534]

kronael:
- CLAUDE.md: comprehensive repo guide (build, architecture, all components, data sources)
- .claude/skills/marinade-sam-bond/SKILL.md: protocol context, settlement types, ecosystem repos
- README.md: fix Node.js 16->20, add missing packages/crates to repo structure, fix typos
- settlement-pipelines/README.md: add fund-settlement/merge-stakes, fix pipeline count 4->6

PR #451, opened 2026-06-16
Add Mixpanel skill-usage tracking hook

MAKSMAINADE:
## What
A `PostToolUse` hook (matched on the `Skill` tool) that records a Mixpanel `skill_used` event when this plugin's skills (`marinade-sam-bond`, `marinade-ecosystem`) are invoked — so we can measure whether the skills are actually used.

## Design
- **Event:** `PostToolUse` / matcher `Skill` — fires once per skill invocation (lowest-noise signal that still identifies *which* skill ran). `Stop`/`SessionEnd` were considered but can't tell which skill was used.
- **Fire-and-forget:** the POST is detached; it never blocks the session.
- **Safe by default:** silent no-op unless `MIXPANEL_TOKEN` is set.
- **Privacy:** sends only `skill`, `session_id`, and event metadata — **no** filesystem paths, prompts, or user content.

## Files
- `plugins/validator-bonds/hooks/hooks.json` — registers the hook
- `plugins/validator-bonds/scripts/track-mixpanel.sh` — detached wrapper; auto-loads `MIXPANEL_TOKEN` from a gitignored `.env` via `CLAUDE_PROJECT_DIR`
- `plugins/validator-bonds/scripts/mixpanel_track.py` — builds + POSTs the event
- `plugins/validator-bonds/hooks/README.md` — what it does, how to enable, privacy
- `.env.example` + `.gitignore` (`.env`) — token config without committing secrets

## Config
Set a Mixpanel **project token** (write-only ingestion token, not an API secret): `export MIXPANEL_TOKEN=...` or copy `.env.example` → `.env`.

## Tested
`mixpanel http=200 body=1` (accepted) via the verbose path; wrapper returns instantly (non-blocking).

> Note: stacked on `skill-audit-and-manifests` (the branch that adds the plugin). Base that branch first, or retarget to wherever the plugin lands.

### DIFF
--- a/plugins/validator-bonds/hooks/hooks.json
+++ b/plugins/validator-bonds/hooks/hooks.json
@@ -1,0 +1,15 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Skill",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "${CLAUDE_PLUGIN_ROOT}/scripts/track-mixpanel.sh"
+          }
+        ]
+      }
+    ]
+  }
+}
--- a/plugins/validator-bonds/scripts/track-mixpanel.sh
+++ b/plugins/validator-bonds/scripts/track-mixpanel.sh
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+# Fire-and-forget Mixpanel usage tracker for the validator-bonds skills.
+#
+# Wired as a PostToolUse hook on the `Skill` tool (see ../hooks/hooks.json).
+# Reads the hook event JSON on stdin and records a `skill_used` event in
+# Mixpanel. Never blocks the session: the network call is detached and this
+# script returns immediately with exit 0.
+#
+# Requires:  MIXPANEL_TOKEN env var (your Mixpanel project token).
+#            If unset, this is a silent no-op.
+#
+# Test:      echo '{"session_id":"t1","tool_name":"Skill",
+#              "tool_input":{"skill":"marinade-sam-bond"},
+#              "hook_event_name":"PostToolUse","cwd":"/tmp"}' \
+#              | MIXPANEL_TOKEN=xxx ./track-mixpanel.sh
+
+INPUT="$(cat)"
+
+# If the token isn't already in the environment, load it from a gitignored
+# .env at the project root (so you don't have to export it in your shell).
+# An explicit MIXPANEL_TOKEN in the environment always takes precedence.
+if [ -z "${MIXPANEL_TOKEN:-}" ] && [ -f "${CLAUDE_PROJECT_DIR:-.}/.env" ]; then
+  set -a; . "${CLAUDE_PROJECT_DIR:-.}/.env"; set +a
+fi
+
+[ -z "${MIXPANEL_TOKEN:-}" ] && exit 0
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Detach the poster so a slow/failing Mixpanel call never stalls the session.
+printf '%s' "$INPUT" | MIXPANEL_TOKEN="$MIXPANEL_TOKEN" \
+  nohup python3 "$DIR/mixpanel_track.py" >/dev/null 2>&1 &
+
+exit 0
--- a/plugins/validator-bonds/skills/marinade-sam-bond/SKILL.md
+++ b/plugins/validator-bonds/skills/marinade-sam-bond/SKILL.md
@@ -1,0 +1,73 @@
+---
+name: marinade-sam-bond
+description: Validator Bonds protocol internals — settlement types, SAM auction, bond collateral, PSR, epoch lifecycle. NOT for ecosystem navigation or issue filing (use marinade-ecosystem).
+when_to_use: CPMPE, PMPE, totalPmpe, PSR, SAM auction, ValidatorBond, SettlementReason, ProtectedEvent, BidTooLowPenalty, BlacklistPenalty, BondRiskFee, InstitutionalPayout, fund_bond, withdraw_request, init_withdraw_request, claim_withdraw_request, merkle settlement, bid-distribution, settlement-config.yaml, programs/validator-bonds/, settlement-distributions/, settlement-pipelines/, packages/validator-bonds-*, mSOL stakers, native staking, Select stakers, institutional stakers, bond collateral, clearing price, winningTotalPmpe, validator bid, dynamic commission, dynamic bids, minimum bond balance, minBondBalance, 7 SOL, program ID, vBoNdEvzMrSai7is21XgVYik65mqtaKXuSdMBJ1xkW4
+---
+
+# Validator Bonds Protocol Context
+
+Validators post SOL bonds as collateral to compete for Marinade's delegated stake via SAM (Stake Auction Marketplace). Bonds guarantee mSOL holders earn at minimum network-average rewards (PSR), and provide a 50bps APY guarantee for institutional/Select stakers -- underperformance triggers automatic compensation from the bond.
+
+**Two staker populations:** mSOL holders (liquid staking, bidding settlements) and institutional/Select stakers (native staking, 50bps APY guarantee).
+
+## Settlement Types
+
+Top-level `SettlementReason` variants (`settlement-common/src/settlement_collection.rs`): `Bidding`, `BidTooLowPenalty`, `BlacklistPenalty`, `BondRiskFee`, `InstitutionalPayout`, and `ProtectedEvent(...)` which wraps a protected-event sub-kind.
+
+| Type                  | SettlementReason          | Trigger                                 | Funder                                     | Recipient                     |
+| --------------------- | ------------------------- | --------------------------------------- | ------------------------------------------ | ----------------------------- |
+| Bidding               | `Bidding`                 | Validator wins auction, owes bid amount | ValidatorBond                              | mSOL stakers                  |
+| BidTooLowPenalty      | `BidTooLowPenalty`        | Bid below minimum threshold             | ValidatorBond                              | Stakers + Marinade/DAO fee    |
+| BlacklistPenalty      | `BlacklistPenalty`        | Blacklisted (sandwich, slow slots)      | ValidatorBond                              | Stakers                       |
+| BondRiskFee           | `BondRiskFee`             | Bond risk premium (scoring-calculated)  | ValidatorBond                              | Stakers                       |
+| DowntimeRevenueImpact | `ProtectedEvent` sub-kind | Fewer credits than expected             | ValidatorBond (0-50%) / Marinade (50-100%) | Stakers                       |
+| CommissionSamIncrease | `ProtectedEvent` sub-kind | Commission raised above declared bid    | ValidatorBond                              | Stakers (with markup penalty) |
+| InstitutionalPayout   | `InstitutionalPayout`     | Select program APY settlement           | ValidatorBond                              | Institutional stakers         |
+
+`ProtectedEvent` (`settlement-common/src/protected_events.rs`) also contains legacy `CommissionIncrease` and `LowCredits` (v1, no longer emitted).
+
+## Epoch Lifecycle
+
+1. **Epoch X**: Validators bid (static CPMPE or dynamic commission, since Jan 2026), SAM allocates stake
+2. **Epoch X+1**: Off-chain pipelines calculate charges from epoch X performance
+3. Merkle trees generated, Settlement accounts created on-chain
+4. Bond stake accounts fund settlements (deactivated)
+5. **Claiming window** (~4 epochs): stakers prove merkle membership, claim rewards
+6. Expired settlements closed, unclaimed funds return to bond
+
+## Key Concepts
+
+- **CPMPE** -- lamports per 1000 SOL per epoch, the validator's fixed bid price
+- **Clearing price** -- `winningTotalPmpe`: PMPE of the last validator group to receive stake in the auction
+- **Program ID** -- `vBoNdEvzMrSai7is21XgVYik65mqtaKXuSdMBJ1xkW4`
+- **Min bond balance** -- 7 SOL (production `minBondBalance`); below this, stake cap is reduced, not eligibility revoked
+- **Bond authority** -- `bond.authority` field or validator identity can sign
+- **fund_bond** transfers stake ownership to bonds PDA; recovery via withdraw request (3-epoch lockup)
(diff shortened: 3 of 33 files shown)


kronael:
Can we make it simpler?  Just a single .py script, please?

kronael:
Also maybe, if you could rebase to the current #360 .  Mostly should be ok to use the skills and plugin from there.

kronael (commented):
🤖 claude: Verified against Claude Code docs. Four issues found, one critical.

**[Critical] PostToolUse hook never fires for the `Skill` tool — see inline comment on `hooks.json:5`.**

**Skill content bugs (SKILL_AUDIT.md calls these out but doesn't fix them):**
- `BidTooLowPenalty` trigger description is wrong (inline on SKILL.md:20)
- Stake cap behavior below 7 SOL floor is wrong (inline on SKILL.md:43)
- `PriorityFee` settlement type is missing (inline on SKILL.md:15)

**Nit — `CLAUDE_EVAL=1` env var has no documented effect:**
`runner.ts` sets `CLAUDE_EVAL: '1'` in the subprocess env. Checked https://code.claude.com/docs/en/env-vars — this variable is not listed and has no recognized effect in Claude Code. Remove it or replace with a real guard (e.g. `--no-prompt-caching`).

**Nit — `plugins/validator-bonds/evals/report/` is not gitignored:**
The runner writes eval output to `./report/<tag>/`. That directory is already untracked in the working tree (visible in `git status`). Add `plugins/validator-bonds/evals/report/` to `.gitignore`.

plugins/validator-bonds/hooks/hooks.json
```diff
@@ -0,0 +1,15 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Skill",
```
kronael, line 5: 🤖 claude: **[Critical] PostToolUse hooks do not fire for the `Skill` tool.**

Verified against https://code.claude.com/docs/en/hooks: the `Skill` tool is handled as internal prompt expansion and bypasses the tool-execution/hook pipeline. A `PostToolUse` matcher on `"Skill"` produces zero events. The shell script and `mixpanel_track.py` are never reached.

Two paths forward:
1. **Wait for upstream support** — GH issue #43630 tracks this; it's currently closed as not-planned.
2. **Use `Stop`/`SessionEnd` hook instead** — fires at end of turn; loses per-invocation granularity but does fire. You'd need to embed which skills ran in the hook payload or track per-session rather than per-invocation.

Until the hook mechanism is fixed, this PR ships a no-op.

plugins/validator-bonds/scripts/track-mixpanel.sh
```diff
@@ -0,0 +1,34 @@
...
+if [ -z "${MIXPANEL_TOKEN:-}" ] && [ -f "${CLAUDE_PROJECT_DIR:-.}/.env" ]; then
+  set -a; . "${CLAUDE_PROJECT_DIR:-.}/.env"; set +a
+fi
+
+[ -z "${MIXPANEL_TOKEN:-}" ] && exit 0
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Detach the poster so a slow/failing Mixpanel call never stalls the session.
+printf '%s' "$INPUT" | MIXPANEL_TOKEN="$MIXPANEL_TOKEN" \
```
kronael, line 31: 🤖 claude: **Nit — the bash wrapper can be eliminated; Python can detach itself.**

The only reason this `.sh` file exists is to do `nohup ... &` so the HTTP call doesn't block. Python can do the same with `subprocess.Popen(start_new_session=True)`, re-execing itself:

```python
import os, sys, subprocess

if not os.environ.get('_TRACK_WORKER'):
    # Launcher: re-exec self as detached worker
    data = sys.stdin.buffer.read()
    proc = subprocess.Popen(
        [sys.executable, __file__],
        env={**os.environ, '_TRACK_WORKER': '1'},
        stdin=subprocess.PIPE,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
        start_new_session=True,
    )
    proc.stdin.write(data)
    proc.stdin.close()
    sys.exit(0)

# Worker: proceed with token check, filtering, HTTP call
# ... (rest of mixpanel_track.py) ...
```

This collapses two files into one, removes the bash dependency, and makes the hook command simply `"command": "${CLAUDE_PLUGIN_ROOT}/scripts/mixpanel_track.py"` directly. (Moot until the hooks.json issue above is resolved, but worth doing when that is fixed.)

plugins/validator-bonds/skills/marinade-sam-bond/SKILL.md
```diff
@@ -0,0 +1,73 @@
...
+
+# Validator Bonds Protocol Context
+
+Validators post SOL bonds as collateral to compete for Marinade's delegated stake via SAM (Stake Auction Marketplace). Bonds guarantee mSOL holders earn at minimum network-average rewards (PSR), and provide a 50bps APY guarantee for institutional/Select stakers -- underperformance triggers automatic compensation from the bond.
+
+**Two staker populations:** mSOL holders (liquid staking, bidding settlements) and institutional/Select stakers (native staking, 50bps APY guarantee).
+
+## Settlement Types
+
+Top-level `SettlementReason` variants (`settlement-common/src/settlement_collection.rs`): `Bidding`, `BidTooLowPenalty`, `BlacklistPenalty`, `BondRiskFee`, `InstitutionalPayout`, and `ProtectedEvent(...)` which wraps a protected-event sub-kind.
```
kronael, line 15: 🤖 claude: **`PriorityFee` is a 7th `SettlementReason` variant missing here and from the table below.**

Verified: `settlement-distributions/settlement-common/src/settlement_collection.rs:32`. Confirmed live in epoch data: `BLX5PkLh…` had 12 PriorityFee charges (0.35 SOL). An operator seeing unexpected charges of this type won't find them explained in the skill. SKILL_AUDIT.md TODO-3 flags this as "should fix".

plugins/validator-bonds/skills/marinade-sam-bond/SKILL.md
```diff
@@ -0,0 +1,73 @@
...
+
+| Type                  | SettlementReason          | Trigger                                 | Funder                                     | Recipient                     |
+| --------------------- | ------------------------- | --------------------------------------- | ------------------------------------------ | ----------------------------- |
+| Bidding               | `Bidding`                 | Validator wins auction, owes bid amount | ValidatorBond                              | mSOL stakers                  |
+| BidTooLowPenalty      | `BidTooLowPenalty`        | Bid below minimum threshold             | ValidatorBond                              | Stakers + Marinade/DAO fee    |
```
kronael, line 20: 🤖 claude: **Wrong trigger — verified against `ds-sam/calculations.ts:248`.**

"Bid below minimum threshold" is not what fires this. The real trigger: **validator lowers its bid compared to the previous epoch**. The upstream comment reads "penalizes validators who reduce their commitment compared to the last epochs."

The eval case `cases/bid-too-low-penalty.yaml` had it right (`facts: [reduces, previous epoch]`); the skill doesn't match its own eval. SKILL_AUDIT.md TODO-2 calls this a must-fix.

Suggested:
```
| BidTooLowPenalty | `BidTooLowPenalty` | Bid reduced vs previous epoch | ValidatorBond | Stakers + Marinade/DAO fee |
```

plugins/validator-bonds/skills/marinade-sam-bond/SKILL.md
```diff
@@ -0,0 +1,73 @@
...
+4. Bond stake accounts fund settlements (deactivated)
+5. **Claiming window** (~4 epochs): stakers prove merkle membership, claim rewards
+6. Expired settlements closed, unclaimed funds return to bond
+
+## Key Concepts
+
+- **CPMPE** -- lamports per 1000 SOL per epoch, the validator's fixed bid price
+- **Clearing price** -- `winningTotalPmpe`: PMPE of the last validator group to receive stake in the auction
+- **Program ID** -- `vBoNdEvzMrSai7is21XgVYik65mqtaKXuSdMBJ1xkW4`
+- **Min bond balance** -- 7 SOL (production `minBondBalance`); below this, stake cap is reduced, not eligibility revoked
```
kronael, line 43: 🤖 claude: **"reduced, not eligibility revoked" is false — verified against `ds-sam/constraints.ts:275` (`clipBondStakeCap`).**

Below **80% of the 7 SOL floor (< 5.6 SOL)**, the stake cap is set to **0** — full revocation. Confirmed live: `DeEpSdaw…` at 0.673 SOL gets zero stake. SKILL_AUDIT.md TODO-1 names this a must-fix.

An operator trusting this line ("stake cap is reduced") will think they still get some delegation when they actually get none.

Suggested: `below 5.6 SOL (80% of floor), stake cap is revoked to 0; from 5.6–7 SOL it is reduced proportionally`

plugins/validator-bonds/hooks/hooks.json
kronael, line 5: 🤖 claude: **[Critical] PostToolUse hooks do not fire for the `Skill` tool.**

Verified against https://code.claude.com/docs/en/hooks: the `Skill` tool is handled as internal prompt expansion and bypasses the tool-execution/hook pipeline. A `PostToolUse` matcher on `"Skill"` produces zero events. The shell script and `mixpanel_track.py` are never reached — the feature is a no-op as written.

Two paths forward:
1. **`Stop`/`SessionEnd` hook** — fires at end of turn; loses per-invocation granularity but does fire. Track which skills ran via session state.
2. **Wait for upstream support** — GH issue anthropics/claude-code#43630 tracks Skill PostToolUse; currently closed as not-planned.

plugins/validator-bonds/scripts/track-mixpanel.sh
kronael, line 31: 🤖 claude: **Nit — the bash wrapper can be collapsed into one Python file.**

The only job of this `.sh` is to `nohup ... &`. Python can do the same by re-execing itself with `subprocess.Popen(start_new_session=True)`:

```python
import os, sys, subprocess

if not os.environ.get("_TRACK_WORKER"):
    data = sys.stdin.buffer.read()
    proc = subprocess.Popen(
        [sys.executable, __file__],
        env={**os.environ, "_TRACK_WORKER": "1"},
        stdin=subprocess.PIPE,
        stdout=subprocess.DEVNULL,
        stderr=subprocess.DEVNULL,
        start_new_session=True,   # nohup equivalent
    )
    proc.stdin.write(data)
    proc.stdin.close()
    sys.exit(0)

# Worker path: token check, filtering, HTTP call (rest of mixpanel_track.py) ...
```

This removes the bash dependency and the two-file split. The hook command would become `"${CLAUDE_PLUGIN_ROOT}/scripts/mixpanel_track.py"` directly. (Moot until the `hooks.json` issue is resolved.)

plugins/validator-bonds/skills/marinade-sam-bond/SKILL.md
kronael, line 15: 🤖 claude: **`PriorityFee` is a 7th `SettlementReason` variant missing here and from the table below.**

Verified: `settlement-distributions/settlement-common/src/settlement_collection.rs:32`. Confirmed live: `BLX5PkLh…` had 12 PriorityFee charges (0.35 SOL) in epoch data. Operators seeing unexpected charges of this type won't find them explained. SKILL_AUDIT.md TODO-3 flags this as "should fix".

plugins/validator-bonds/skills/marinade-sam-bond/SKILL.md
kronael, line 20: 🤖 claude: **Wrong trigger — verified against `ds-sam/calculations.ts:248`.**

"Bid below minimum threshold" is not what fires this. The real trigger: **validator lowers its bid compared to the previous epoch**. The upstream comment reads "penalizes validators who reduce their commitment compared to the last epochs."

The eval case `cases/bid-too-low-penalty.yaml` had it right (`facts: [reduces, previous epoch]`); the skill doesn't match its own eval. SKILL_AUDIT.md TODO-2 calls this a must-fix.

Suggested: `| BidTooLowPenalty | ... | Bid reduced vs previous epoch | ...`

plugins/validator-bonds/skills/marinade-sam-bond/SKILL.md
kronael, line 43: 🤖 claude: **"reduced, not eligibility revoked" is false — verified against `ds-sam/constraints.ts:275` (`clipBondStakeCap`).**

Below **80% of the 7 SOL floor (< 5.6 SOL)**, the stake cap is set to **0** — full revocation, not reduction. Confirmed live: `DeEpSdaw…` at 0.673 SOL gets zero stake. SKILL_AUDIT.md TODO-1 names this a must-fix.

An operator trusting this line will think they still get some delegation when they actually get none.

Suggested: `below 5.6 SOL (80% of floor), stake cap is revoked to 0; from 5.6–7 SOL it is reduced proportionally`

kronael:
@MAKSMAINADE I added some comments.  Would not change the SKILLS at all, however, if its not needed.

MAKSMAINADE:
Superseded by feat/skill-usage-tracking (stacked on [#360](https://github.com/marinade-finance/validator-bonds/pull/360)) — closing this in favor of it. Recap of how that PR addresses the review here:

[Critical] PostToolUse/Skill no-op — confirmed against the [hooks docs](https://code.claude.com/docs/en/hooks) and [anthropics/claude-code#43630](https://github.com/anthropics/claude-code/issues/43630): the Skill tool is prompt expansion and never enters the tool/hook pipeline. Reworked to fire on the Stop hook and scan transcript_path for Skill tool_use blocks (which are recorded), emitting one skill_used event per new invocation, deduped by tool_use id. Verified against a real transcript (http=200 body=1, dedup works). A code comment points to [#43630](https://github.com/anthropics/claude-code/issues/43630) to swap back to PostToolUse when it ships.

Nit — eliminate the bash wrapper — done: collapsed to a single .py that re-execs itself detached via subprocess.Popen(start_new_session=True), exactly as suggested. No more .sh.

Nit — CLAUDE_EVAL=1 — removed; the subprocess still inherits process.env by default.

Nit — evals/report/ not gitignored — added to .gitignore.

Skill-content bugs (BidTooLowPenalty trigger / stake-cap-below-floor / missing PriorityFee) — already corrected in [#360](https://github.com/marinade-finance/validator-bonds/pull/360), which the new PR builds on.

Thanks for the catch on the critical one — it was indeed shipping a no-op.

PR #452, opened 2026-06-16
Add Mixpanel skill-usage tracking (Stop hook) + eval nits

MAKSMAINADE:
Stacked on #360 (`20260312_skill`). Supersedes #451.

Two commits:

### 1. Skill-usage tracking hook
Records a Mixpanel `skill_used` event when a plugin skill runs.
- **Not PostToolUse:** the Skill tool is prompt expansion and never fires Pre/PostToolUse hooks ([docs](https://code.claude.com/docs/en/hooks), [#43630](https://github.com/anthropics/claude-code/issues/43630)). Instead this fires on **Stop**, scans `transcript_path` for `Skill` `tool_use` blocks, and emits one event per new invocation (deduped by `tool_use` id).
- Single self-contained `.py`, re-execs itself detached (`subprocess.Popen(start_new_session=True)`) so it never blocks the session.
- Verified against a real transcript: detects the skill, posts (`http=200 body=1`), dedups on re-scan, returns ~60ms.
- Mixpanel *project* token (write-only, public-safe) hardcoded; override via `MIXPANEL_TOKEN`. Sends only `skill`, `session_id`, `tool_use_id`, event metadata — no paths/prompts/content.

### 2. Eval-harness nits (from #451 review)
- `eval.ts`: drop `CLAUDE_EVAL=1` (not a real Claude Code env var, no effect).
- `.gitignore`: ignore `plugins/validator-bonds/evals/report/`.

All four #451 review comments are resolved here (the three SKILL.md content bugs were already fixed in #360).


plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,151 @@
...
+does NOT fire Pre/PostToolUse hooks (anthropics/claude-code#43630). But each Skill
+invocation IS recorded in the session transcript as a `tool_use` block. So on every
+Stop we scan `transcript_path` for Skill invocations and emit one `skill_used` event
+per new invocation, deduped by the immutable `tool_use` id across the per-turn Stop
+calls. Switch to a PostToolUse/`Skill` matcher once #43630 ships.
+
+Token: MIXPANEL_PROJECT_TOKEN is a Mixpanel *project* token — write-only ingestion,
+safe to ship public. Override with the MIXPANEL_TOKEN env var.
+
+The launcher re-execs itself detached (`start_new_session`) so a slow Mixpanel call
```
kronael: The re-execution should have a strict timeout ... e.g. claude has it, but if it hangs, we need to prevent detached zombies.

plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,151 @@
...
+    changed = False
+    for tid, skill in skill_invocations(tpath):
+        if skill in TRACKED_SKILLS and tid not in seen:
+            post(sid, skill, tid)
+            seen.add(tid)
+            changed = True
+    if changed:
+        try:
+            with open(sp, "w") as fh:
+                json.dump(sorted(seen), fh)
```
kronael: Could this be done in mixpanel by constructing the dedup_id in a way such that no seen file would be needed?

plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,151 @@
...
+        except OSError:
+            pass
+
+
+raw = sys.stdin.read()
```
kronael: It should be normally wrapped in 

if __name__ == '__main__': ...

otherwise its king of... ugly

plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,151 @@
...
+        env={**os.environ, "_TRACK_WORKER": "1"},
+        stdin=subprocess.PIPE,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+        start_new_session=True,
+    )
+    p.stdin.write(raw.encode())
+    p.stdin.close()
+except Exception:
+    run(raw)  # fallback: run inline if we can't detach
```
kronael: Can this fire? If not, would not have it and would just exit...

plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,151 @@
...
+            "skill": skill,
+            "tool_use_id": tid,
+            "hook_event": "Stop",
+            "source": "claude-code",
+        },
+    }
+    if VERBOSE:
+        print(f"emit skill_used skill={skill} tid={tid}")
+    data = urllib.parse.urlencode({"data": json.dumps(payload)}).encode()
+    req = urllib.request.Request(
```
kronael, line 104: Can more events be posted in a single round?  If so, would opt to send all at once from the whole run.

kronael (commented):
Thank you .  Posting some comments.

MAKSMAINADE:
Thanks for the review — all 5 addressed in 120a957:

Strict timeout / no detached zombies — the detached worker now arms signal.alarm(10), on top of the existing 5s timeout on the Mixpanel POST. Hard self-kill, nothing lingers.
Dedup without a seen-file — switched to Mixpanel $insert_id set to the invocation's tool_use id. Re-scanning the transcript on later Stop calls is now deduped server-side, so the local state file is gone entirely.
if __name__ == "__main__" — entry logic moved into main() and wrapped.
The except: run() fallback — removed. If we can't detach, it just exits now; a blocking synchronous POST would defeat the point of detaching.
Batching — collects every skill invocation from the run and sends them in a single /track request (chunked at 50, Mixpanel's per-request cap).

plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,157 @@
...
+        with open(transcript_path, encoding="utf-8") as fh:
+            for line in fh:
+                if '"Skill"' not in line:
+                    continue
+                try:
+                    rec = json.loads(line)
+                except Exception:
+                    continue
+                msg = rec.get("message") or {}
+                content = msg.get("content") if isinstance(msg, dict) else None
```
kronael: else [] would simplify the check below possibly.

plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,157 @@
...
+    if not TOKEN:
+        return
+    try:
+        ev = json.loads(raw or "{}")
+    except Exception:
+        return
+    if not ev.get("transcript_path"):
+        return
+    events = build_events(ev)
+    for i in range(0, len(events), MIXPANEL_BATCH):
```
kronael: would use a lib: https://docs.python.org/3/library/itertools.html#itertools.batched  (normally I wouldn't but its a stdlib so imo better than shipping our own)

plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,157 @@
...
+                        "$insert_id": tid,  # server-side dedup key
+                        "time": now,
+                        "skill": skill,
+                        "tool_use_id": tid,
+                        "hook_event": "Stop",
+                        "source": "claude-code",
+                    },
+                }
+            )
+    return events
```
kronael: nit: might also yield per event

plugins/validator-bonds/scripts/track-mixpanel.py
kronael: Then you'd just batch the iterator here and not the list.

plugins/validator-bonds/scripts/track-mixpanel.py
```diff
@@ -0,0 +1,157 @@
...
+
+
+def main():
+    raw = sys.stdin.read()
+
+    # Worker (and verbose/testing) path: do the work, with a hard self-timeout so
+    # a detached worker can never hang around.
+    if VERBOSE or os.environ.get("_TRACK_WORKER"):
+        try:
+            signal.alarm(WORKER_TIMEOUT_SECONDS)
```
kronael, line 144: Is there a handler setup for the signal, however?  E.G. will anything happen once it fires?

kronael (approved):
some nits.. mostly would see if the signal is really there to kill the stray process in case it times out

kronael (changes requested):
🤖 claude:

**Security pass on the Stop hook — request changes: disclose the telemetry. The code itself is fine.**

No data leak: `track-mixpanel.py` parses the transcript locally and sends only the skill name (the 4 marinade skills) + opaque ids (`session_id`, `tool_use_id`) + timestamp — no prompts, content, file paths, or bond/staker data. The `if '"Skill"' not in line: continue` prefilter means non-skill message bodies are never even deserialized. Injection-safe (list-form `Popen`, `json.loads`, urlencoded body, single pinned host).

**Blocker:** the telemetry is undisclosed — it POSTs to Mixpanel on every Stop, with no mention in the README and no documented opt-out.

**Requested change:** add a `## Telemetry` section to `plugins/validator-bonds/README.md` (after Install) covering:
- **Sent:** skill name + `session_id` + `tool_use_id` + timestamp.
- **Never sent:** prompts, responses, file paths, file contents, conversation/repo data.
- **Opt out:** empty `MIXPANEL_TOKEN` (keeps the skills, kills only the telemetry); or `claude plugin disable|uninstall validator-bonds@marinade-finance` to drop it entirely. Repo / `--plugin-dir` installs can delete `hooks/hooks.json`; marketplace installs can't remove just the hook (cached copy is restored on update).

That's the only change requested.

MAKSMAINADE:
Added the ## Telemetry section to plugins/validator-bonds/README.md (after Install) in d40cc2ff — covers exactly what you asked:

Sent: skill name + session_id + tool_use_id + timestamp.
Never sent: prompts, responses, file paths, file contents, conversation/repo/bond/staker data.
Opt out: empty MIXPANEL_TOKEN (keeps skills, kills telemetry); or claude plugin disable|uninstall; repo/--plugin-dir installs can delete hooks/hooks.json (marketplace installs can't remove just the hook — cached copy restored on update, so use the token opt-out / disable).

PR #454, opened 2026-06-22
[feature-gate] pipeline fix on fund processing with 1 SOL limit [GEN-8174]

ochaloup:
Pipeline fixes to adapt with 1 SOL minimum delegation is required onchain change
https://explorer.solana.com/address/STk5Xj8hdAx3sTzmtJ3QysKkq6X2A3yj73JtxttiRyk

### DIFF
--- a/settlement-pipelines/src/bin/fund_settlement.rs
+++ b/settlement-pipelines/src/bin/fund_settlement.rs
@@ -285,11 +285,11 @@ async fn prepare_funding(
             continue;
         }
 
-        let settlement_amount_funded = funded_to_settlement_stakes
-            .get(&settlement_record.settlement_address)
-            .map_or(0, |(lamports_in_accounts, funded_accounts)|
-                // Amount funded has to be calculated without the minimal stake amount that is not part of the claim
-                *lamports_in_accounts - (funded_accounts.len() as u64 * minimal_stake_lamports));
+        // derive from on-chain fields; stake-account balances under-reserve for stakes funded before a minimum_stake_lamports increase and underflowed
+        let settlement_amount_funded = settlement_record
+            .settlement_account
+            .as_ref()
+            .map_or(0, |s| s.lamports_funded.saturating_sub(s.lamports_claimed));
         let amount_to_fund = settlement_record.settlement_account.as_ref().map_or(
             settlement_record.max_total_claim_sum,
             |settlement| {
--- a/settlement-pipelines/src/bin/fund_settlement.rs
+++ b/settlement-pipelines/src/bin/fund_settlement.rs
@@ -56,7 +56,7 @@ use validator_bonds_common::config::get_config;
 use validator_bonds_common::constants::find_event_authority;
 use validator_bonds_common::stake_accounts::{
     collect_stake_accounts, get_clock, get_stake_history, obtain_delegated_stake_accounts,
-    obtain_funded_stake_accounts_for_settlement, CollectedStakeAccount, CollectedStakeAccounts,
+    CollectedStakeAccount, CollectedStakeAccounts,
 };
 
 #[derive(Parser, Debug)]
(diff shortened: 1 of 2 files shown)


settlement-pipelines/src/bin/fund_settlement.rs
```diff
@@ -445,11 +411,68 @@ async fn prepare_funding(
...
+                            continue;
+                        }
+                        validator_bonds_funders.push(SettlementFunderValidatorBond {
                             stake_account_to_fund: stake_account_address,
-                        },
-                    ));
+                        });
+                    }
 
                     match funding_lamports_accumulated
```
kronael, line 477: 🤖 claude: Funded-amount report can overstate funding. All three `match` arms report from `funding_lamports_accumulated` (the pre-skip total of every selected account) or `amount_to_fund`, but the set actually pushed to `validator_bonds_funders` (what `fund_settlements` submits) excludes the merge destination when `fund_destination == false` (L428) and any non-mergeable with `lamports <= minimal_stake_lamports` (L459). Worst case: zero `FundSettlement` instructions are emitted, yet `add_funded_settlement` records a nonzero funded amount. Reporting-only — feeds no on-chain decision, and on-chain `lamports_funded` self-corrects on the next run. Suggest accumulating the actually-funded lamports (`destination_merged_lamports - minimal_stake_lamports` when funded, plus each pushed non-mergeable's `lamports - minimal_stake_lamports`, capped by remaining need) and reporting that across all three arms.

settlement-pipelines/src/bin/fund_settlement.rs
```diff
@@ -390,7 +370,10 @@ async fn prepare_funding(
 
                 // for the found and fitting stake accounts: taking first one and trying to merge other ones into it
                 let stake_account_to_fund: Option<(FundBondStakeAccount, StakeAccountStateType)> =
-                    if stake_accounts_to_fund.is_empty() || funding_lamports_accumulated == 0 {
+                    // below the on-chain minimal stake size the program cannot split nor fund-whole (would underflow at fund_settlement.rs:199), so skip
```
kronael: 🤖 claude: Stale line ref — the underflowing subtraction `get_lamports() - stake_account_min_size` is at `fund_settlement.rs:200`, not `:199` (199 is the explanatory comment added in this branch). The other two comment refs in this diff correctly point at `:200`.

settlement-pipelines/src/bin/fund_settlement.rs
```diff
@@ -482,12 +505,13 @@ async fn prepare_funding(
                             // the rest of lamports will be available in the split stake account
                             // and that can be used as a source for funding of next settlement when vote account is part of multiple settlements
                             // WARN: this REQUIRES that the merge stake transactions are executed in sequence!
-                            let lamports_available_after_split = funding_lamports_accumulated
+                            let lamports_available_after_split = destination_merged_lamports
```
kronael, line 499: 🤖 claude: Pre-existing latent issue (not a regression in this PR): `lamports_available_after_split` ignores `split_stake_rent_exempt`. On-chain the split lands with `fund_split_leftover = get_lamports() - amount_needed - split_stake_rent_exempt` — i.e. `split_stake_rent_exempt` lamports *less* than this estimate. If the estimate equals exactly `minimal_stake_lamports`, the real split account holds `minimum_stake_lamports < stake_account_min_size`, so reusing it for the next settlement would underflow at `fund_settlement.rs:200`. The threshold (`>= minimal_stake_lamports`) is identical to pre-PR code; this PR narrows the gap (base switched from `funding_lamports_accumulated` to `destination_merged_lamports`), it does not widen it. Flagging for awareness.

kronael (dismissed):
Posting what Claude found, but all seem very deep and clearly something either not a problem or nitpick.  I am diving deeper, but don't why we should not merge.

ochaloup:
@kronael Thanks a lot for the review. I updated the code with some of the findings from your review. Can I ask you for a re-review? :pray:

settlement-pipelines/src/bin/fund_settlement.rs
```diff
@@ -469,6 +473,8 @@ async fn prepare_funding(
                             )).with_vote(settlement_record.vote_account_address).add();
                             continue;
                         }
+                        published_funding_lamports +=
```
kronael, line 470: Just verify please, if the non-mergeable are really not accounted above twice.  I'd think not, but I only just undertood most of this yesterday.

settlement-pipelines/src/bin/fund_settlement.rs
```diff
@@ -519,11 +516,16 @@ async fn prepare_funding(
                                     state: destination_stake_state,
                                 });
                             }
-                            reporting
-                                .reportable
-                                .add_funded_settlement(settlement_record, amount_to_fund);
                         }
                     }
+                    // report what was actually published to funding (capped by the need known from the merkle tree), not the planned total
```
kronael: nit: I mostly think the comments will drift and cause damage later-on.  The names are memorable and I wouldn't trust the comment anyway, but an LLM might.

kronael (dismissed):

settlement-pipelines/src/bin/fund_settlement.rs
ochaloup: thanks :pray:  -> checked. not double counted

kronael (approved):

PR #456, opened 2026-06-22
[skill] update penalty commission config [trivial]

ochaloup:
a follow-up to https://github.com/marinade-finance/validator-bonds/pull/453
WDYTY?


kronael:
There will be a followup removing the set-ability altogether...

kronael (approved):

PR #465, opened 2026-06-25
[feat] Add SwapSettlementStake instruction

kronael:
An operator-gated instruction that atomically swaps a settlement's delegated stake for a user-provided undelegated one of equal value. The user's stake is delegated to the settlement's validator and instantly deactivated — effective stake zero, so it is immediately claimable, while staying a delegated stake of that validator so its unclaimed remainder reaps to the bond at close. The user receives the settlement's original delegated stake.

This lets a settlement become claimable without waiting an epoch for the bond stake to deactivate. The operator co-signs each swap so it can enforce AML/KYC off-chain; it touches no claim accounting (orthogonal to ClaimSettlementV2). Includes the SDK builder and bankrun coverage of the swap, its guards, and the full fund/swap/claim/close flow.


kronael:
🤖 claude: **Status — `SwapSettlementStake` + same-epoch payout pipeline**

The instruction and the pipeline wiring are implemented and committed.

**`SwapSettlementStake` instruction** — atomically swaps a settlement's still-deactivating bond stake for a reserve stake of equal value, (re)delegated to the settlement's validator and instantly deactivated. Effective stake 0 → claimable this epoch, yet still a delegated stake of the validator → reaps to the bond at close (`ResetStake`). Operator-gated (AML/KYC). Accepts a fresh `Initialized` stake or a fully-deactivated reserve stake (reserve reuse).

**`fund-settlements` swap pass** — after funding, emits a swap for each `Bidding` settlement: provisions a `marinade_wallet` reserve stake of equal size (split off the grand reserve if present, else create fresh for cold start), skips if the reserve can't cover it, idempotent via the swapped stake's `activation_epoch`. Reuses the deactivated bond stakes received from prior swaps as the next reserve source — so the grand-account lifecycle closes without an explicit merge step.

**Test coverage (bankrun, passing):** swap happy-path, delegated-input guard (6079), unequal-value guard (6080), deactivated-input reuse, wrong-operator guard (6003), and an e2e `fund → swap → claim same-epoch → close → reset` flow.

**Adding two more integration tests:**

- **(A) bankrun two-cycle reserve-reuse** — cycle 1 swaps (marinade receives the bond's deactivated stake) → cycle 2 reuses that stake as the swap input for a new settlement. Proves the grand-account lifecycle end-to-end.
- **(B) test-validator pipeline test** — extends `pipelineSettlement.spec.ts` with a `Bidding` settlement + `marinade_wallet` funding, running the actual `fund-settlement` binary doing the swap and asserting same-epoch claimability.

kronael:
🤖 claude: **Update — swap eligibility corrected, and both integration tests landed.**

Correction to the status above. The swap pass **cannot** key off `SettlementReason::Bidding`:

- `fund-settlement` loads only merkle-tree collections and the loader hard-codes `reason: None` (the `reason: Some(..)` path is dead-code legacy) — so the original `reason == Bidding` filter never matched and the pass was inert.
- The unified pipeline groups settlements by `(vote_account, funder)` and **merges all their claims into one merkle tree / one on-chain settlement**, so a single bond-funded settlement's stake backs bidding + bond-funded-PSR + penalty claims together — reason isn't separable per settlement anyway.

The pass now gates on **`funder == ValidatorBond`**: exactly the settlements funded by a split bond stake that deactivates and lags. Marinade-funded settlements get a fresh `Initialized` (claimable) stake and never lag, so they're skipped. The swap also runs **only when `--marinade-wallet` (the reserve) is explicitly provided**, so a normal `fund-settlement` run without it is unchanged.

Both integration tests are in and passing:

- **`swapSettlementStakeReuse.spec.ts`** (bankrun) — two-cycle reserve reuse: cycle 1 swaps a settlement's deactivating bond stake out to the reserve wallet; after it cools to fully deactivated, cycle 2 reuses that very stake as the swap input and claims same-epoch. Proves the grand-account lifecycle end-to-end.
- **`pipelineSwapSettlement.spec.ts`** (test-validator) — drives the real `fund-settlement` binary with `--marinade-wallet`, then asserts on-chain that the reserve received the bond stake and the settlement-owned stake is the reserve stake re-delegated (and deactivated) this same epoch.

PR #471, opened 2026-07-14
[api] update api to get more info about auction when eventing is run [GEN-8188]

ochaloup:
Providing more data from bond event processing to the API, which the CLI can use to provide guidance to validators.

A follow-up task for implementing CLI guidance.

Once https://github.com/marinade-finance/ds-sam/pull/121 is published as an npm package, this task will use the Bonds API to retrieve the data needed for the CLI to provide guidance to validators.

## Summary by CodeRabbit

* **New Features**
  * Added `GET /bonds/bidding/auction` to return bidding auction metadata and participating validators.
  * Auction context is now persisted and served as part of the bidding auction flow.
* **Bug Fixes**
  * Improved auction metadata serialization by sanitizing non-finite numbers and preventing JSON serialization issues from cyclic data.
  * Auction metadata persistence now refreshes correctly on updates.
* **Documentation**
  * Updated OpenAPI documentation to include the new bidding-auction endpoint and schema.
* **Tests**
  * Expanded Jest coverage for relay serialization and validator-to-state conversion (auction fields).
  * Increased settlement test timeout for reliability.
* **Chores**
  * Updated relevant package/dependency versions and enabled additional database feature flags.

### DIFF
--- a/packages/bonds-eventing/src/run-auction.ts
+++ b/packages/bonds-eventing/src/run-auction.ts
@@ -7,6 +7,7 @@ import {
   loadSamConfig,
 } from '@marinade.finance/ds-sam-sdk'
 
+import type { AuctionMeta } from './calc-relay'
 import type { EventingConfig } from './types'
 import type { LoggerWrapper } from '@marinade.finance/ts-common'
 
@@ -16,7 +17,7 @@ export async function runAuction(
 ): Promise<{
   validators: AuctionValidator[]
   epoch: number
-  winningTotalPmpe: number
+  meta: AuctionMeta
 }> {
   const productionConfig = await loadSamConfig()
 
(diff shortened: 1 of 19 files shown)


packages/bonds-eventing/src/run-auction.ts
```diff
@@ -48,6 +49,19 @@ export async function runAuction(
   return {
     validators,
     epoch,
-    winningTotalPmpe: result.winningTotalPmpe,
+    meta: {
```
kronael, line 52: just thinking here... could we save "more" and don't think about it that much?  E.G. `{ result, cfg }`?  The we would have some place less to make changes when things change.

kronael (dismissed):

packages/bonds-eventing/src/run-auction.ts
ochaloup: I was thinking about it, and it makes quite a lot of sense. I’m in favor of changing it.

ochaloup:
@kronael, thanks for the review. I updated based on your idea. Can I ask for a re-review :pray:

kronael (approved):

PR #478, opened 2026-07-24
[GEN-8423] serve verified validators list from API

ilmari-h:
Adds endpoint `/validators/verified`

List of verified validators is a config file in ops-infra

## Summary by CodeRabbit

* **New Features**
  * Added `--verified-validators-config` API startup option to load a verified-validator list from a YAML file (missing → warning and empty list; load failure → startup fails).
  * Added a rate-limited `GET /validators/verified` endpoint to return the configured list, including OpenAPI documentation.
* **Documentation**
  * Updated the API “start” command example and refreshed the `verified-validators.yaml` example.
* **Bug Fixes**
  * YAML entries are now strictly validated as Solana public keys; invalid or missing expected keys cause the config load to fail rather than being skipped.


api/src/repositories/verified_validators.rs
```diff
@@ -0,0 +1,34 @@
...
+    pub verified_validators: Vec<String>,
+}
+
+pub fn load_verified_validators(path: &str) -> anyhow::Result<Vec<String>> {
+    let config: VerifiedValidatorsConfig = read_from_yaml_file(&path)?;
+    Ok(validate(config))
+}
+
+// Keeps only entries that parse as a valid pubkey; invalid ones are logged and dropped so a
+// single typo in the config cannot take the endpoint down.
```
ochaloup: If I understand correctly, the list of validators is loaded on startup and then kept in memory. I believe that with an invalid pubkey in the config, the service should refuse to start rather than only report a warning. Kubernetes will keep routing traffic through the old pods until the new one is fully ready, so we have time to fix the config properly.

api/README.md
```diff
@@ -52,7 +52,8 @@ export DB='validator-bonds'
 export PG_SSLROOTCERT="/tmp/${DB}-postgres-root-cert.pem"
 export POSTGRES_URL="postgresql://${DB}:${DB}@localhost:5444/${DB}"
 cargo run --bin api -- --postgres-url "$POSTGRES_URL" \
-  --postgres-ssl-root-cert "$PG_SSLROOTCERT"
+  --postgres-ssl-root-cert "$PG_SSLROOTCERT" \
+  --verified-validators-config ./verified-validators.yaml
```
ochaloup: Possibly to add somewhere a valid example of `./verified-validators.yaml` - like `./verified-validators.yaml.example`?

api/src/repositories/verified_validators.rs
```diff
@@ -0,0 +1,34 @@
...
+use solana_sdk::pubkey::Pubkey;
+use std::str::FromStr;
+
+#[derive(Debug, Deserialize)]
+pub struct VerifiedValidatorsConfig {
+    pub verified_validators: Vec<String>,
+}
+
+pub fn load_verified_validators(path: &str) -> anyhow::Result<Vec<String>> {
+    let config: VerifiedValidatorsConfig = read_from_yaml_file(&path)?;
```
ochaloup, line 14: Maybe to dedup (and sort?)

api/src/repositories/verified_validators.rs
```diff
@@ -0,0 +1,34 @@
+use serde::Deserialize;
+use settlement_common::utils::read_from_yaml_file;
+use solana_sdk::pubkey::Pubkey;
+use std::str::FromStr;
+
+#[derive(Debug, Deserialize)]
+pub struct VerifiedValidatorsConfig {
+    pub verified_validators: Vec<String>,
```
ochaloup, line 10: ApiDoc registers a Pubkey string schema via the PubkeyScheme modifier, but the response field is a plain Vec<String>

api/src/bin/api.rs
```diff
@@ -52,10 +56,24 @@ async fn main() -> anyhow::Result<()> {
...
+    let verified_validators = match &params.verified_validators_config {
+        Some(path) => {
+            verified_validators_repo::load_verified_validators(path).unwrap_or_else(|err| {
+                error!("Failed to load verified validators config from {path}: {err}");
+                vec![]
+            })
+        }
+        None => {
+            error!("Verified validators config not provided, will serve an empty list.");
+            vec![]
```
ochaloup, line 68: Don't swallow a load failure when the path was given explicitly.

Agreeing with CodeRabbit here. Right now a missing or malformed YAML produces the same 200 as a legitimately empty list, so a typo rolls out clean and silently empties the endpoint with no way for a consumer to tell. Suggest propagating the error when `--verified-validators-config` is passed, and keeping the empty-list default only for the branch where the flag is omitted. Given the rollout is a
  rolling restart, failing startup means the old pod keeps serving the last-good list — which is the behaviour we want.

That also lets the omitted-flag branch drop from `error!` to `warn!`: no config provided is a deliberate opt-out, not an error, and at error level it will fire on every boot until the ops-infra change lands.

The per-entry tolerance in `validate()` is a good call and I'd keep it as is.

ochaloup (commented):
I have few comments to consider.

api/src/repositories/verified_validators.rs
ilmari-h: fixed

api/src/bin/api.rs
ilmari-h: fixed

api/README.md
ilmari-h: added

api/src/repositories/verified_validators.rs
```diff
@@ -0,0 +1,28 @@
...
+// endpoint. Entries are deduplicated and sorted for a stable response.
+fn validate(config: VerifiedValidatorsConfig) -> anyhow::Result<Vec<String>> {
+    let mut verified = BTreeSet::new();
+    for entry in config.verified_validators {
+        let pubkey = Pubkey::from_str(&entry)
+            .with_context(|| format!("Invalid verified validator vote account '{entry}'"))?;
+        verified.insert(pubkey.to_string());
+    }
+    Ok(verified.into_iter().collect())
+}
```
ochaloup, line 28: 🤖 claude: This flipped lenient → fail-fast and added dedup+sort, but nothing locks it in: `cargo test -p api` runs 0 tests and the crate has no `#[cfg(test)]` module at all.

Risk: someone hits a crashloop from one bad key in the ConfigMap, restores `unwrap_or_else(|_| vec![])` to unblock the deploy, CI stays green, and the endpoint quietly serves an incomplete list behind a 200 that no consumer can distinguish from a short list.

Worth a small `mod tests` here: invalid entry → `Err`, duplicate → len 1, unsorted input → sorted output, missing `verified_validators` key → `Err`.

api/src/repositories/verified_validators.rs
```diff
@@ -0,0 +1,28 @@
...
+    pub verified_validators: Vec<String>,
+}
+
+pub fn load_verified_validators(path: &str) -> anyhow::Result<Vec<String>> {
+    let config: VerifiedValidatorsConfig = read_from_yaml_file(&path)?;
+    validate(config)
+}
+
+// Fails on any invalid pubkey so a single typo aborts startup rather than silently emptying the
+// endpoint. Entries are deduplicated and sorted for a stable response.
```
ochaloup: 🤖 claude: Nit — the second sentence restates what `BTreeSet` + `into_iter().collect()` already show two lines down. The first sentence carries the why; I'd keep just that.

api/README.md
```diff
@@ -52,7 +52,8 @@ export DB='validator-bonds'
 export PG_SSLROOTCERT="/tmp/${DB}-postgres-root-cert.pem"
 export POSTGRES_URL="postgresql://${DB}:${DB}@localhost:5444/${DB}"
 cargo run --bin api -- --postgres-url "$POSTGRES_URL" \
-  --postgres-ssl-root-cert "$PG_SSLROOTCERT"
+  --postgres-ssl-root-cert "$PG_SSLROOTCERT" \
+  --verified-validators-config ./verified-validators.yaml  # see verified-validators.yaml.example
```
ochaloup, line 56: 🤖 claude: The relative paths don't match the cwd the rest of this README assumes. Line 25 iterates `./migrations/*.sql` and `migrations/` sits at the repo root, so from there `verified-validators.yaml.example` doesn't exist (it's `api/verified-validators.yaml.example`), and nothing tells the reader to create `./verified-validators.yaml` — which is now a fatal startup error.

This also covers CodeRabbit's thread above: its suggested `cp ./verified-validators.yaml.example ./verified-validators.yaml` breaks from the repo root for the same reason.

```bash
cp api/verified-validators.yaml.example verified-validators.yaml
cargo run --bin api -- --postgres-url "$POSTGRES_URL" \
  --postgres-ssl-root-cert "$PG_SSLROOTCERT" \
  --verified-validators-config ./verified-validators.yaml
```

ochaloup (commented):
A few follow-up comments—feel free to decline them if you wish. I believe this can be approved once the review points are closed or resolved.


More comments below

- Optional: `validate()` aborts on the first bad key, where the old code logged all of them. N typos = N deploy cycles.
- Optional follow-up: nothing catches a dangling OpenAPI `$ref`. `9e772e7` shipped `#/components/schemas/solana_sdk.pubkey.Pubkey` past clippy, tests and CI; a test asserting every `$ref` resolves would cover all future `value_type` uses.

api/verified-validators.yaml.example
```diff
@@ -0,0 +1,3 @@
+verified_validators:
```
ochaloup, line 1: 🤖 claude: The real file this gets copied to isn't ignored — `git check-ignore api/verified-validators.yaml` exits 1 and the root `.gitignore` has no matching pattern at either location. A dev following the README gets a dirty `git status`, and an `git add -A` on an unrelated change commits their local verified-validators list.

Worth adding next to the `# Local` block in `.gitignore`:

```
verified-validators.yaml
api/verified-validators.yaml
```

api/src/repositories/verified_validators.rs
ilmari-h: fixed

api/src/repositories/verified_validators.rs
ilmari-h: added a test

ochaloup (approved):

PR #492, opened 2026-08-11
[GEN-8055][cli] notification info fixes and update definition

ochaloup:
* Fixing the CLI to force the update to be only for major/minor versions
* Updating the package.json versions for release
* Readme update

## Summary by CodeRabbit

- **New Features**
  - Added CLI version checking that permits patch updates with a warning and installation guidance.
  - Major and minor version updates are blocked until the CLI is upgraded.
  - Added commands for managing bond notification subscriptions and viewing notifications.
  - `show-bond` now supports withdrawal-request addresses.

- **Documentation**
  - Documented Select bond funding thresholds and Telegram/email notification workflows.

- **Chores**
  - Updated CLI packages to version 2.6.0.

### DIFF
--- a/packages/validator-bonds-cli-core/__tests__/npmRegistry.spec.ts
+++ b/packages/validator-bonds-cli-core/__tests__/npmRegistry.spec.ts
@@ -1,7 +1,7 @@
 import {
+  checkCliVersion,
   compareVersions,
   fetchLatestVersionInNpmRegistry,
-  requireLatestCliVersion,
 } from '../src/npmRegistry'
 
 import type { Logger } from 'pino'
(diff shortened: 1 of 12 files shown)


packages/validator-bonds-cli-core/__tests__/npmRegistry.spec.ts
```diff
@@ -344,31 +344,98 @@ describe('fetchLatestVersionInNpmRegistry', () => {
...
+    await expect(
+      checkCliVersion(mockLogger, registryUrl, '2.6.0'),
+    ).resolves.toBeUndefined()
+    await expect(
+      checkCliVersion(mockLogger, registryUrl, '2.7.0'),
+    ).resolves.toBeUndefined()
+    expect(stderrSpy).not.toHaveBeenCalled()
+  })
+
+  it('lets an unreachable registry pass', async () => {
```
martinkrecek, line 429: (non-blocking suggestion)
Maybe it would be worth it to also print a warning here, because (assuming that the CLI user isn't aware of the failure based on the assertion - I might be wrong here) it can happen that they hit some compatibility issues and aren't aware that the fetch failed, so they keep trying with an outdated version.

(I consider this to be a total edge case - using the CLI with NPM down or no internet access - so it's not a blocker for merging)

martinkrecek (approved):

packages/validator-bonds-cli-core/__tests__/npmRegistry.spec.ts
ochaloup: That's intentional. We don't want to print any information about this to the user and make them face errors around registry unavailability that are unrelated to what they're working on on-chain.

PR #496, opened 2026-08-20
[GEN-8771] Dependencies version bump [maintenance]

ochaloup:
Updating the versions to be up-to-date

## Summary by CodeRabbit

* **New Features**
  * Bond initialization commands now support an optional bond authority.
  * Validator and bond reporting preserves additional delegation and stake-cap details.
  * CLI errors now include complete chained-cause information.

* **Bug Fixes**
  * Improved handling of auction, withdrawal, stake, and settlement data.
  * Network errors are classified more accurately, including wrapped failures.
  * Fee simulations now exclude bidding settlements with missing details.
  * Stake lockup status correctly reflects expiration and timing conditions.

* **Maintenance**
  * Updated package versions and supporting tooling for improved compatibility and reliability.


packages/validator-bonds-sdk/src/web3.js/stakeAccount.ts
```diff
@@ -88,19 +88,17 @@ async function parseStakeAccountData(
...
     isCoolingDown: deactivationEpoch ? !deactivationEpoch.eq(U64_MAX) : false,
     isLockedUp:
       lockup !== undefined &&
-      lockup.custodian &&
-      lockup.custodian !== undefined &&
       lockup.custodian !== PublicKey.default &&
-      (lockup?.epoch.gt(currentEpoch) ||
-        lockup?.unixTimestamp.gt(currentTimestamp)),
+      (lockup.epoch.gt(currentEpoch) ||
+        lockup.unixTimestamp.gt(currentTimestamp)),
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed, but by dropping the custodian check entirely rather than switching to `.equals()` — agave's `Lockup::is_in_force(clock, None)` (solana-stake-interface `src/state.rs`) decides lockup purely from `unix_timestamp`/`epoch`, and the custodian field only exempts a *signing* custodian, so it never decides whether a lockup is in force. `.equals(PublicKey.default)` would have reported an un-liftable lockup (default custodian + future epoch) as unlocked and fed that stake to bond funding via `parseNotLocked`; a bankrun test now pins all three cases (default custodian, real custodian, no lockup).

packages/bonds-eventing/src/index.ts
```diff
@@ -43,7 +47,9 @@ program.parseAsync(process.argv).then(
     logger.debug({ resolution: 'Success', args: process.argv })
   },
   (err: Error) => {
-    logger.error(err.message)
+    logger.error(
+      err instanceof ErrorWithCause ? err.messageWithCause() : err.message,
+    )
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — all three top-level rejection handlers (`bonds-eventing`, `validator-bonds-sanity-check`, `mainCommand`) now take `(err: unknown)` and fall back to `String(err)`. Widening the parameter type was the load-bearing part: left as `Error` the new guard is unreachable to the type checker, and `mainCommand` additionally only calls `translateKnownError` for real `Error`s since that helper casts its input internally.

ochaloup:
🤖 claude: Recording decisions on the three items that have no inline thread to reply to.

- **`scripts/fee-annotation.ts` — restore `&& s.details !== null`**: declining. The `Bidding` variant is `details: BidDetails` (non-null) and the producer at `bid-distribution/src/generators/bidding.rs:618` always emits `Some(SettlementDetails::Bidding(..))`, so the case cannot occur; the suggested guard also would not catch it, since the Rust field is `Option` with `skip_serializing_if`, meaning an absent value is *omitted* and reads back as `undefined`, which `!== null` passes.
- **Ticket reference in PR title**: declining for now — this is a dependency-maintenance PR with no ticket assigned.
- **Docstring coverage 20.83% vs 80%**: declining. It conflicts with this repo's documented policy (`CLAUDE.md`: comments "ZERO by default", "never multi-line"), so satisfying the gate would require exactly the doc blocks the policy forbids.

packages/validator-bonds-sdk/__tests__/bankrun/solanaStake.spec.ts
```diff
@@ -948,4 +952,41 @@ describe('Solana stake account behavior verification', () => {
...
+      (await getStakeAccount(provider, defaultCustodian, currentEpoch))
+        .isLockedUp,
+    ).toBe(true)
+    expect(
+      (await getStakeAccount(provider, realCustodian, currentEpoch)).isLockedUp,
+    ).toBe(true)
+    expect(
+      (await getStakeAccount(provider, noLockup, currentEpoch)).isLockedUp,
+    ).toBe(false)
+  })
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — added a `Lockup(futureTimestamp, 0, PublicKey.default)` case asserting `isLockedUp === true`; all three existing cases left `unixTimestamp` at `0`, so the `unixTimestamp.gt(currentTimestamp)` branch was genuinely uncovered. One detail worth recording: `parseStakeAccountData` compares against wall-clock `Date.now()` rather than the bankrun clock, so the case uses a real future unix time — and I confirmed it fails (`Expected: true, Received: false`) when that branch alone is removed, so it covers the intended condition rather than the epoch one.

packages/bonds-eventing/src/calc-relay.ts
```diff
@@ -52,8 +52,10 @@ export function toCalcValidator(v: AuctionValidator): Record<string, unknown> {
     idealUnprotectedReserve: v.idealUnprotectedReserve,
     bondGoodForNEpochs: v.bondGoodForNEpochs,
     unstakePriority: v.unstakePriority,
+    maxBondDelegation: v.maxBondDelegation,
+    bondSamStakeCapSol: v.bondSamStakeCapSol,
     auctionStake: { marinadeSamTargetSol: v.auctionStake.marinadeSamTargetSol },
-    bondForcedUndelegation: { value: v.bondForcedUndelegation?.value ?? null },
+    bondForcedUndelegation: { value: v.bondForcedUndelegation.value },
```
ilmari-h: This is the one guard in `toCalcValidator` that wasn't incidental — every other field was written unguarded, while this one got both `?.` **and** an explicit `?? null`. Reads like someone hit an undefined here rather than defensive boilerplate.

`bondForcedUndelegation` is per-validator state that only means something for validators actually being force-undelegated. ds-sam-sdk 0.2.0 declares it required, but that could just be an optimistic type — worth checking the assignment in the SDK is unconditional before trusting it.

If it isn't: `TypeError: Cannot read properties of undefined (reading 'value')` on the first unaffected validator. `toCalcValidator` is called per validator inside the delta loop, so the whole eventing run dies and no bond notifications go out for the epoch.

If unconditional assignment isn't confirmable, restoring `?.value ?? null` costs nothing — `jsonSafe` passes null straight through.

scripts/fee-annotation.ts
```diff
@@ -47,7 +47,7 @@ async function main() {
   }
 
   const bids = settlements.filter(
-    (s): s is BidSettlement => s.reason === 'Bidding' && s.details !== null,
+    (s): s is BidSettlement => s.reason === 'Bidding',
```
ilmari-h: This is a type predicate, so the compiler takes it on faith. Dropping the `details` check is correct against the union in `settlement-utils.ts:27` (the `Bidding` arm declares `details: BidDetails`) — but that union is a hand-written mirror of a Rust struct describing *today's* JSON, and this script reads settlement JSON off disk, including old epochs via `regression-test-settlements.sh`. Per CLAUDE.md the schema changed at epoch 928, with ≤927 in the older format.

A `Bidding` entry carrying `details: null` now passes the filter, gets typed as having details, and hits line 53:

```
b.details.total_marinade_active_stake
→ TypeError: Cannot read properties of null
```

Previously that entry was dropped and totals came from the rest.

Suggest keeping `&& s.details !== null` with a `// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition` noting the JSON is external and pre-928 epochs predate the schema — same pattern this PR already uses well in `utils.ts:64` and `show.spec.ts`.

Unrelated but while here: the generated PR summary describes this as "Fee simulations include all bidding settlements, even when details are unavailable." It's the opposite — when details are unavailable it now throws. Probably shouldn't stay in the description.

scripts/simulate-fee.ts
```diff
@@ -307,7 +307,7 @@ for (let epoch = epochStart; epoch <= epochEnd; epoch++) {
       adj_min_fee_bps?: number
     }
     const bidSettlements = settlements.filter(
-      (s): s is BidSettlement => s.reason === 'Bidding' && s.details !== null,
+      (s): s is BidSettlement => s.reason === 'Bidding',
```
ilmari-h: Same as `fee-annotation.ts:50` — predicate now trusts `reason` alone, and `bidDetails` feeds straight into `d.total_marinade_active_stake` / `d.total_marinade_stakers_rewards` reducers below. A pre-928 `Bidding` entry with `details: null` throws instead of being skipped. This one matters more since it's the script driven across epoch ranges.

packages/validator-bonds-cli-core/src/commands/mainCommand.ts
```diff
@@ -189,12 +189,12 @@ export function launchCliProgram({
...
       // Argument parsers like parsePubkey return Promise<PublicKey>; unwrap so
       // we inspect the resolved value, not the pending Promise.
-      const arg = await Promise.resolve(action.processedArgs?.[0]).catch(
+      const arg = await Promise.resolve(action.processedArgs[0]).catch(
         () => undefined,
       )
       const account = arg instanceof PublicKey ? arg.toBase58() : undefined
       const { accountField } = getProgramTelemetryFields(action)
-      const walletPubkey = walletInterface.publicKey?.toBase58()
+      const walletPubkey = walletInterface.publicKey.toBase58()
```
ilmari-h: The nullability question here is secondary to the structural one: this whole telemetry block runs inside `preAction` with no try/catch. A throw rejects the hook promise, so Commander never runs the command body — user types `fund-bond`, gets a stack trace, no transaction attempted, killed by the analytics path.

The line right above already knows this:

```ts
const arg = await Promise.resolve(action.processedArgs[0]).catch(() => undefined)
```

That `.catch` exists so a rejected arg parser can't take down the command. Same instinct, one line up.

On whether it can fire: `parseWalletFromOpts` takes `printOnly` as an argument, so it has a distinct no-real-keypair path; whether that yields a real pubkey or a placeholder is a web3js-1x 5.1.2 internal. The old `?.` suggests the original author wasn't sure either.

Rather than chase that, suggest wrapping the block in `try { … } catch { /* telemetry must never fail a command */ }`. Kills the whole class of failure and makes the `?.` question moot.

packages/bonds-eventing/src/calc-relay.ts
ilmari-h, line 58: ### 🤖 Verify ds-sam-sdk assigns this unconditionally

Only guard in `toCalcValidator` that wasn't incidental — every other field was already unguarded, this one had `?.` *and* `?? null`. `bondForcedUndelegation` is state that only applies to validators actually being force-undelegated, so the 0.2.0 type declaring it required may be optimistic.

If it is, this throws on the first unaffected validator and takes down the whole eventing run — `toCalcValidator` is per-validator inside the delta loop.

Unless the SDK's assignment is confirmed unconditional, restore the guard; `jsonSafe` passes null through:

```suggestion
    bondForcedUndelegation: { value: v.bondForcedUndelegation?.value ?? null },
```

scripts/fee-annotation.ts
ilmari-h: ### 🤖 Predicate trusts a union that pre-928 JSON violates

Correct against the union in `settlement-utils.ts:27`, but that union is a hand-written mirror of a Rust struct describing *current* JSON. This script reads settlement files off disk, including ≤927 epochs via `regression-test-settlements.sh`, which predate the `details` schema.

A `Bidding` entry with `details: null` now passes, gets typed as having details, and throws at line 53 (`b.details.total_marinade_active_stake`). Previously it was skipped.

```suggestion
    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- settlement JSON is external; pre-928 epochs predate the details schema
    (s): s is BidSettlement => s.reason === 'Bidding' && s.details !== null,
```

scripts/simulate-fee.ts
ilmari-h: ### 🤖 Same predicate, driven across epoch ranges

Same issue as `fee-annotation.ts:50`, and `bidDetails` feeds the `d.total_marinade_active_stake` / `d.total_marinade_stakers_rewards` reducers below.

```suggestion
      // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- settlement JSON is external; pre-928 epochs predate the details schema
      (s): s is BidSettlement => s.reason === 'Bidding' && s.details !== null,
```

packages/validator-bonds-cli-core/src/commands/mainCommand.ts
ilmari-h, line 197: ### 🤖 Telemetry throw kills the command

Nullability aside, this block runs in `preAction` with no try/catch. A throw rejects the hook promise and Commander never runs the command body — user runs `fund-bond`, gets a stack trace, no transaction attempted. Whether `parseWalletFromOpts` yields a real pubkey on the `--print-only` path is a web3js-1x 5.1.2 internal.

Wrap the block instead of chasing the type:

```ts
try {
  // … existing telemetry block …
} catch {
  // telemetry must never fail a command
}
```

scripts/fee-annotation.ts
ochaloup: 🤖 claude: Fixed, but with `!= null` rather than `!== null` — epochs 750/860/880 omit the `details` key entirely (Rust writes it with `skip_serializing_if = "Option::is_none"`), so the strict check let them through and the pre-branch code crashed identically. Epoch 880 now prints “No Bidding settlements found” and epoch 918 is unchanged.

scripts/simulate-fee.ts
ochaloup: 🤖 claude: Fixed the same way — `s.details != null`, since the key is absent rather than null on pre-~890 epochs and the strict form would not have caught it. `-c 880` now falls through to the `no data` continue instead of throwing in the `activeStake` reducer.

packages/bonds-eventing/src/calc-relay.ts
ochaloup: 🤖 claude: Declining — ds-sam-sdk 0.2.0 `transformValidators` spreads `validatorAggDefaults()` or `ineligibleValidatorAggDefaults()` on all five of its return paths, and both initialise `bondForcedUndelegation: { coef: 0, base: 0, value: 0 }`. The conditional assignment in `auction.js` overwrites that default for affected validators, it does not create the field, and 0.1.1 declared it required too.

packages/validator-bonds-cli-core/src/commands/mainCommand.ts
ochaloup: 🤖 claude: Declining — every wallet `parseWalletFromOpts` can return has an assigned `publicKey` (the `--print-only`/`show` fallback is `NullWallet`, whose `publicKey` is `PublicKey.default`), and commander initialises `processedArgs` to `[]`, so nothing in the block can throw. The try/catch is fair hardening on its own merit, but it guards nothing today and is unrelated to the dependency bump.

scripts/fee-annotation.ts
```diff
@@ -47,7 +47,8 @@ async function main() {
   }
 
   const bids = settlements.filter(
-    (s): s is BidSettlement => s.reason === 'Bidding' && s.details !== null,
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- settlement JSON is external; epochs before ~890 omit `details` on Bidding entries
+    (s): s is BidSettlement => s.reason === 'Bidding' && s.details != null,
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Took the first half — `isFeeSettlement` now also requires `s.details != null`, so its `s is FeeSettlement` predicate stops lying on legacy input instead of relying on the caller's earlier `bids.length === 0` return; epoch 918 and 970 (47 PriorityFee records) produce byte-identical output. Declining the shebang: the only caller is `pnpm exec tsx scripts/fee-annotation.ts` in `.buildkite/prepare-bid-distribution.yml`, which is what the usage text already says, the file is not executable so a shebang is inert, and it uses no Bun API. No regression test — every epoch sampled from 710 to 1019 is homogeneous and `bidding.rs` emits `Some(..)` unconditionally, so the mixed record shape it would pin has never existed.

ilmari-h (approved):

PR #497, opened 2026-08-20
[GEN-8772] adopt the shared verifiable-build workflow

janlegner:
Builds through the org's shared workflow; all 12 checks green. `verify-onchain` is deliberately false because `build.rs` bakes `env!("GIT_REV")` into `security_txt` and a caller cannot supply it — `cargo-args` is static, GitHub expressions cannot slice `github.sha` to a short rev, and the deployed tag doesn't contain this file — so it records a hash rather than reporting a harness artefact as drift. Tag trigger is `contract-v*`; the 56 `v*` tags are npm/SDK releases that don't change the program.

Measured baseline: `5c7e0fa636f2f839…`, 1,113,689 bytes, slot 410,576,659. `Anchor.toml` says `solana_version = 2.3.1` while the lockfile pins 2.3.0 — worth reconciling separately.

⚠️ Blocked on marinade-finance/.github#6.

---
_Generated by [Claude Code](https://claude.ai/code)_

## Summary by CodeRabbit

* **New Features**
  * Added a manually triggered verifiable build workflow.
  * Verifiable builds now run automatically for contract version tags and relevant pull requests.
  * Builds are configured for reproducible validation without on-chain verification.

### DIFF
--- a/.github/workflows/verifiable-build.yml
+++ b/.github/workflows/verifiable-build.yml
@@ -5,6 +5,12 @@ on:
   push:
     tags:
       - 'contract-v*'
+  pull_request:
+    paths:
+      - 'programs/**'
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'Anchor.toml'
 
 permissions:
   contents: read
--- a/.github/workflows/verifiable-build.yml
+++ b/.github/workflows/verifiable-build.yml
@@ -1,0 +1,58 @@
+# Calls the org's shared verifiable-build workflow, which builds the program
+# reproducibly and compares the result against the bytecode deployed at the
+# program ID in Anchor.toml's [programs.mainnet] table.
+#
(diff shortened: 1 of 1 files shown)


.github/workflows/verifiable-build.yml
```diff
@@ -0,0 +1,22 @@
+name: Verifiable Build
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - 'contract-v*'
```
ochaloup, line 7: Please add similarly what is here - run only for changes at path `/programs`
https://github.com/marinade-finance/marinade-config/pull/3/changes#diff-c2398c0e0a7026d887755e9b7ec8c8072e6cce2368c1255b22d502a13d519a5fR9-R14

ochaloup (changes requested):
I was asked for a review, but the PR is in draft state. I'm not sure whether that was intentional.

ochaloup (approved):

PR #499, opened 2026-08-21
[GEN-8245] fix api loading  protected events processing

ochaloup:
On checking the https://github.com/marinade-finance/validator-bonds/pull/491, there seems to be some issues coming from bq pagination. Fixing that mostly + other smaller points.
Connected to here: https://github.com/marinade-finance/ops-infra/pull/2780

---

`/protected-events` and `/v1/protected-events` report the complete event history again, and stop re-rendering it on every request.

- Page the BigQuery read and reject a short result, since a single-page read silently dropped 13,917 rows — epochs 530–732 vanished from the feed after `/v1` widened the query past BigQuery's one-page limit
- Poll a bounded number of times when BigQuery answers `job_complete: false`, so a slow query no longer caches an empty result that reads as "no validator owes a protected event"
- Render both response bodies once per cache refresh instead of per request, so a request is a `Bytes` clone rather than a 64k-record clone plus a 12 MB serde render that one IP could demand 30 times a second
- Add an optional `?from_epoch=` window to both endpoints so the PSR dashboard can pull the recent epochs it needs instead of 489 epochs of history, with the memoized body still served whenever the window drops nothing

## Summary by CodeRabbit

* **New Features**
  * Added optional `from_epoch` filtering to both protected-events endpoints.
  * Improved protected-event response ordering and support for narrowed time windows.
  * Added API documentation for the new query parameter and possible error responses.

* **Bug Fixes**
  * Improved handling of paginated and incomplete data results.
  * Added validation for reported row counts and query completion.
  * Cache refresh failures now retry sooner, improving data availability.

### DIFF
--- a/api/src/context.rs
+++ b/api/src/context.rs
@@ -15,25 +15,37 @@ pub type ProtectedEventsCache = Arc<RwLock<Option<ProtectedEvents>>>;
 /// request. `records` stays for the incremental refresh and for the `from_epoch` windows.
 pub struct ProtectedEvents {
     pub records: Vec<ProtectedEventRecord>,
+    pub min_epoch: Option<u64>,
     pub v1_body: Bytes,
     pub legacy_body: Bytes,
 }
 
 impl ProtectedEvents {
-    pub fn new(records: Vec<ProtectedEventRecord>) -> anyhow::Result<Self> {
+    pub fn new(mut records: Vec<ProtectedEventRecord>) -> anyhow::Result<Self> {
+        // The refresh appends the fresh block to the older one; unsorted, the feed jumps back up.
+        records.sort_by(|left, right| right.epoch.cmp(&left.epoch));
         let legacy = LegacyProtectedEventsResponse {
             protected_events: legacy_projection(&records),
         };
         let legacy_body = Bytes::from(serde_json::to_vec(&legacy)?);
+        let min_epoch = records.iter().map(|record| record.epoch).min();
         let v1 = ProtectedEventsResponse {
             protected_events: records,
         };
         let v1_body = Bytes::from(serde_json::to_vec(&v1)?);
         Ok(Self {
             records: v1.protected_events,
+            min_epoch,
             v1_body,
             legacy_body,
         })
+    }
+
+    /// Whether the window drops a record, so a window reaching the whole feed can answer from the
+    /// rendered body.
+    pub fn narrows(&self, from_epoch: u64) -> bool {
+        self.min_epoch
+            .is_some_and(|min_epoch| from_epoch > min_epoch)
     }
 }
 
--- a/api/src/context.rs
+++ b/api/src/context.rs
@@ -1,12 +1,41 @@
+use axum::body::Bytes;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 use tokio_postgres::Client;
 
-use crate::dto::ProtectedEventRecord;
+use crate::dto::{
+    legacy_projection, LegacyProtectedEventsResponse, ProtectedEventRecord, ProtectedEventsResponse,
+};
 
 /// `None` until a BigQuery fetch has succeeded once. Distinguishes "never loaded" from an epoch
 /// that genuinely settled nothing, which the handlers answer as 500 and 200 respectively.
-pub type ProtectedEventsCache = Arc<RwLock<Option<Vec<ProtectedEventRecord>>>>;
+pub type ProtectedEventsCache = Arc<RwLock<Option<ProtectedEvents>>>;
+
+/// Both endpoints answer the whole feed, so the bodies are rendered per refresh instead of per
+/// request. `records` stays for the incremental refresh and for the `from_epoch` windows.
+pub struct ProtectedEvents {
+    pub records: Vec<ProtectedEventRecord>,
+    pub v1_body: Bytes,
+    pub legacy_body: Bytes,
+}
+
+impl ProtectedEvents {
+    pub fn new(records: Vec<ProtectedEventRecord>) -> anyhow::Result<Self> {
+        let legacy = LegacyProtectedEventsResponse {
+            protected_events: legacy_projection(&records),
+        };
+        let legacy_body = Bytes::from(serde_json::to_vec(&legacy)?);
+        let v1 = ProtectedEventsResponse {
+            protected_events: records,
+        };
+        let v1_body = Bytes::from(serde_json::to_vec(&v1)?);
+        Ok(Self {
+            records: v1.protected_events,
+            v1_body,
+            legacy_body,
+        })
+    }
+}
 
 pub struct Context {
     pub psql_client: Client,
--- a/api/src/handlers/protected_events.rs
+++ b/api/src/handlers/protected_events.rs
@@ -1,32 +1,53 @@
 use crate::{
     context::WrappedContext,
-    dto::{legacy_projection, LegacyProtectedEventRecord, ProtectedEventRecord},
+    dto::{
+        legacy_projection, LegacyProtectedEventsResponse, ProtectedEventRecord,
+        ProtectedEventsResponse,
+    },
     error::AppError,
 };
+use axum::body::Bytes;
 use axum::extract::{Query, State};
-use axum::Json;
+use axum::http::header;
+use axum::response::IntoResponse;
 use serde::{Deserialize, Serialize};
-
-#[derive(Serialize, Debug, utoipa::ToSchema)]
-pub struct ProtectedEventsResponse {
-    protected_events: Vec<ProtectedEventRecord>,
-}
-
-#[derive(Serialize, Debug, utoipa::ToSchema)]
-#[schema(deprecated)]
-pub struct LegacyProtectedEventsResponse {
-    protected_events: Vec<LegacyProtectedEventRecord>,
-}
 
 #[derive(Deserialize, Serialize, Debug, utoipa::IntoParams)]
 #[into_params(parameter_in = Query)]
-pub struct QueryParams {}
+pub struct QueryParams {
+    pub from_epoch: Option<u64>,
+}
+
+fn json_response(body: Bytes) -> impl IntoResponse {
+    ([(header::CONTENT_TYPE, "application/json")], body)
+}
+
+fn render(response: &impl Serialize) -> Result<Bytes, AppError> {
+    let body = serde_json::to_vec(response).map_err(|err| AppError {
+        message: format!("Failed to render the protected events: {err}"),
+    })?;
+    Ok(Bytes::from(body))
+}
+
+fn records_from_epoch(
+    records: &[ProtectedEventRecord],
+    from_epoch: u64,
+) -> Vec<ProtectedEventRecord> {
+    records
+        .iter()
+        .filter(|record| record.epoch >= from_epoch)
+        .cloned()
+        .collect()
+}
 
 #[utoipa::path(
     get,
     tag = "Protected Events",
     operation_id = "List Bid PSR (protected events)",
     path = "/protected-events",
+    params(
+        ("from_epoch" = Option<u64>, Query, description = "Report epochs from this one on, inclusive. Omitted, the whole history is reported."),
+    ),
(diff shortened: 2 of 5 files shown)


api/src/repositories/protected_events.rs
```diff
@@ -33,17 +40,82 @@ async fn get_protected_events(
...
+                polls <= MAX_COMPLETION_POLLS,
+                "BigQuery job {job_id} has not completed after {MAX_COMPLETION_POLLS} polls"
+            );
+        }
+        let results = client
+            .job()
+            .get_query_results(project_id, &job_id, parameters)
+            .await?;
+        rs = ResultSet::new(results.into());
     }
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Declining — that branch is only reachable when `job_complete != true`, and `ResultSet::new` (gcp-bigquery-client 0.17.1, `model/query_response.rs`) sets `row_count = 0` in exactly that case, so `next_row()` pushed nothing and `protected_events` is already empty. The `clear()` would be a no-op, and this repo's CLAUDE.md bans handling for unreachable states.

api/src/handlers/protected_events.rs
```diff
@@ -1,80 +1,270 @@
...
+            Body::Window(LegacyProtectedEventsResponse {
+                protected_events: legacy_projection(records_from_epoch(
+                    &events.records,
+                    from_epoch,
+                )),
+            })
+        }
+        _ => Body::Rendered(events.legacy_body.clone()),
+    }
+}
```
ilmari-h, line 83: ### 🤖 `?from_epoch=` reopens the render-per-request path this PR closes

`narrows()` compares against `min_epoch`, i.e. the *oldest* epoch in the feed — around 530 in production. So `?from_epoch=531` drops one epoch's rows and takes `Body::Window`, which clones the records (line 61) and serde-renders essentially the whole feed, per request.

`routes.rs:29` sets `PUBLIC_RATE_PER_SEC = 30` per IP. That's the same 30 req/s × 12 MB the PR body cites as the thing being fixed — it now just needs one query parameter.

Options, cheapest first:
- serialize by reference (`Vec<&ProtectedEventRecord>`) — kills the clone, not the render;
- `records` is already sorted descending, so `partition_point` instead of the linear `filter` in `records_from_epoch`;
- memoize the windows that matter: bucket `from_epoch` (or a tiny LRU keyed on it), rendered once per refresh. The dashboard pulls a fixed recent window, so one or two entries covers real traffic;
- or clamp the served span and document it, so no window costs more than N epochs of render.

api/src/repositories/protected_events.rs
ilmari-h, line 64: ### 🤖 The paging loop has no overall iteration bound

`MAX_COMPLETION_POLLS` only counts iterations where `parameters.page_token.is_none()` (line 52). Two shapes escape it:

- `job_complete: false` **with** a page token — `next_page` returns `Some` carrying the token, so `polls` never increments;
- a page token BigQuery keeps echoing without advancing.

Either loops forever, and there's no wall-clock bound on the fetch. Because the loop sits before the `sleep` in `spawn_protected_events_cache_updater`, a wedged fetch means the cache silently stops refreshing with nothing logged — the failure looks like nothing at all.

Suggest bounding total iterations as well as token-less polls, or wrapping `get_protected_events` in `tokio::time::timeout` so a wedged read surfaces as the retriable error it is.

api/src/repositories/protected_events.rs
```diff
@@ -33,17 +40,82 @@ async fn get_protected_events(
...
+    Ok(Some((
+        job_id,
+        GetQueryResultsParameters {
+            page_token,
+            // mandatory outside US and EU, and the queried dataset is europe-central2
+            location: job.location,
+            timeout_ms: Some(COMPLETION_POLL_TIMEOUT_MS),
+            ..Default::default()
+        },
+    )))
```
ilmari-h, line 94: ### 🤖 The token-less re-ask relies on an unstated API guarantee

When `job_complete: false`, this re-asks with `page_token: None` and no `start_index`, so the response is read from the top of the result set. That's only safe because the BigQuery API guarantees an incomplete job carries no rows — if a response ever arrives with rows *and* `job_complete: false`, those rows get appended twice.

It does fail safe (`ensure_all_rows_loaded` trips on `loaded > total_rows`), but the error — `Read 127834 of 63917 rows BigQuery reports` — points at the wrong thing and would cost real time to diagnose. `start_index: Some(protected_events.len())` makes the assumption explicit and self-correcting.

api/src/repositories/protected_events.rs
```diff
@@ -77,6 +149,16 @@ fn parse_row(
...
 
+async fn store(cache: &ProtectedEventsCache, records: Vec<ProtectedEventRecord>, stored: &str) {
+    match ProtectedEvents::new(records) {
+        Ok(events) => {
+            *cache.write().await = Some(events);
+            log::info!("{stored}");
+        }
+        Err(err) => log::error!("{stored} failed, the protected events did not render: {err}"),
+    }
+}
```
ilmari-h, line 160: ### 🤖 A render failure still waits out the full refresh interval

`store()` logs and swallows. If `ProtectedEvents::new` fails on a cold cache, the cache stays `None` and both endpoints keep answering 500 — but the updater already took the `Ok(_)` arm and returns `CACHE_UPDATE_INTERVAL` (line 245), so the next attempt is an hour out. The comment on `CACHE_RETRY_INTERVAL` (line 15) states an invariant the code only holds for fetch failures, not render failures.

`serde_json::to_vec` over a `Vec` of plain structs is close to unreachable in practice, so low practical risk — but it's a one-line fix: return `Result<(), _>` from `store` and pick `CACHE_RETRY_INTERVAL` when it fails.

api/src/repositories/protected_events.rs
```diff
@@ -143,20 +230,27 @@ pub fn spawn_protected_events_cache_updater(
...
+                        &protected_events,
+                        merged_protected_events,
+                        "Successfully extended the protected events",
+                    )
+                    .await;
+                    CACHE_UPDATE_INTERVAL
+                }
+                Err(err) => {
+                    log::error!("Failed to get the protected events: {err}");
+                    CACHE_RETRY_INTERVAL
```
ilmari-h, line 249: ### 🤖 The 60 s retry has no backoff, and each attempt is the whole history

On a cold cache `max_loaded_epoch` is 0, so every retry re-runs the full 64k-row union query, and each attempt can spend up to `MAX_COMPLETION_POLLS × COMPLETION_POLL_TIMEOUT_MS` = 300 s polling before failing.

3600 s → 60 s is a 60× increase in query load and cost against BigQuery at exactly the moment BigQuery is already unhealthy, sustained for the length of the outage. Recovering fast from a *transient* failure is the right instinct; the missing piece is a ceiling — back off from `CACHE_RETRY_INTERVAL` up to `CACHE_UPDATE_INTERVAL` (doubling) rather than a flat 60 s.

api/src/handlers/protected_events.rs
ochaloup: 🤖 claude: Fixed — records are epoch-descending, so a window is a prefix of the rendered body; `EpochWindowedBody` keeps a per-epoch byte offset and serves `from_epoch` as a slice, so no request re-renders. That removed `Body<T>` and the clone with it.

api/src/repositories/protected_events.rs
ochaloup: 🤖 claude: Fixed — `fetch_protected_events` wraps the read in a 900 s `tokio::time::timeout` at both call sites, so a non-advancing token surfaces as a logged error on the retry path instead of wedging the updater silently.

api/src/repositories/protected_events.rs
ochaloup: 🤖 claude: Declining — `ResultSet::new` forces `row_count = 0` whenever `job_complete` is not true, so the accumulator is provably empty on that branch and `start_index` would widen `next_page`'s signature for a case the crate cannot produce. `ensure_all_rows_loaded` stays the backstop; CodeRabbit raised and withdrew the same path above.

api/src/repositories/protected_events.rs
ochaloup: 🤖 claude: Fixed — `store` now returns whether the render landed, and the updater picks `CACHE_RETRY_INTERVAL` when it did not, so the comment on that constant holds for render failures too.

api/src/repositories/protected_events.rs
ochaloup: 🤖 claude: Fixed — the retry now doubles per consecutive failure, saturating at `CACHE_UPDATE_INTERVAL` and resetting on success, so a sustained BigQuery outage decays back to hourly instead of holding at 60 s.

api/src/context.rs
```diff
@@ -1,12 +1,53 @@
...
+    pub records: Vec<ProtectedEventRecord>,
+    pub min_epoch: Option<u64>,
+    pub v1_body: Bytes,
+    pub legacy_body: Bytes,
+}
+
+impl ProtectedEvents {
+    pub fn new(mut records: Vec<ProtectedEventRecord>) -> anyhow::Result<Self> {
+        // The refresh appends the fresh block to the older one; unsorted, the feed jumps back up.
+        records.sort_by(|left, right| right.epoch.cmp(&left.epoch));
```
kronael, line 26: can sort_by_key do?

api/src/context.rs
```diff
@@ -1,12 +1,53 @@
...
+};
 
 /// `None` until a BigQuery fetch has succeeded once. Distinguishes "never loaded" from an epoch
 /// that genuinely settled nothing, which the handlers answer as 500 and 200 respectively.
-pub type ProtectedEventsCache = Arc<RwLock<Option<Vec<ProtectedEventRecord>>>>;
+pub type ProtectedEventsCache = Arc<RwLock<Option<ProtectedEvents>>>;
+
+/// Both endpoints answer the whole feed, so the bodies are rendered per refresh instead of per
+/// request. `records` stays for the incremental refresh and for the `from_epoch` windows.
+pub struct ProtectedEvents {
```
kronael, line 16: If the format was json-lines, this could be even pre-rendered for all requests.  But possibly not needed.... using simd_json should speed things up 10x or more imho.

api/src/repositories/protected_events.rs
```diff
@@ -33,17 +40,82 @@ async fn get_protected_events(
...
 
     Ok(protected_events)
 }
 
-fn parse_row(
-    rs: &gcp_bigquery_client::model::query_response::ResultSet,
-) -> anyhow::Result<ProtectedEventRecord> {
+// A query outrunning its request timeout answers `job_complete: false` with no page token, and its
+// rows are then reachable only by re-asking for the job itself.
+fn next_page(rs: &ResultSet) -> anyhow::Result<Option<(String, GetQueryResultsParameters)>> {
```
kronael, line 72: These look pretty nasty... will try to add them to rust-common.

kronael (approved):
pretty nice

api/src/context.rs
ochaloup: 🤖 claude: Yes — now `records.sort_by_key(|record| Reverse(record.epoch))`. Same stable descending order, which the per-epoch byte offsets rely on.

api/src/context.rs
ochaloup: 🤖 claude: Declining both — the feed is already pre-rendered for all requests without a format change: records sort epoch-descending, so every `from_epoch` window is a byte-prefix slice of the rendered body. That also moves rendering to once per refresh rather than per request, which leaves `simd_json` about a tenth of a second an hour to win.

api/src/repositories/protected_events.rs
ochaloup: 🤖 claude: Agreed the paging plumbing belongs in a shared crate — `next_page` and `ensure_all_rows_loaded` only exist because `gcp-bigquery-client` does not page for you. Keeping them here for now so this PR does not pick up a `rust-common` dependency and release; happy to switch over once they land there.

PR #510, opened 2026-09-07
[GEN-8976][rent reduction] split pinned stake reserve from live rent

ochaloup:
## Summary by CodeRabbit

- **Bug Fixes**
  - Stake-account funding and withdrawal validation now consistently enforce the required minimum balance.
  - Settlement workflows use a pinned reserve for stake-account minimums while applying live rent requirements to split accounts and eligible refunds.
  - Improved handling of split-account rent and refund calculations across varying rent scenarios.

- **New Features**
  - Added shared stake-account minimum-balance calculations for consistent CLI and settlement behavior.

- **Tests**
  - Expanded coverage for reduced rent values, insufficient withdrawals, funding thresholds, and settlement balance scenarios.

### DIFF
--- a/packages/validator-bonds-sdk/src/web3.js/stakeAccount.ts
+++ b/packages/validator-bonds-sdk/src/web3.js/stakeAccount.ts
@@ -267,7 +267,7 @@ export async function getRentExemptStake(
   )
 }
 
-// Pinned into every Meta by the stake program, which ignores it and takes the real floor from the Rent sysvar (SIMD-0490); bonds compares against the stored field, so never use the live rent here.
+// The bonds program checks the stored Meta.rent_exempt_reserve, which the stake program still pins at this value (SIMD-0490) even though live rent is now lower (SIMD-0437) — mirror it, never read live rent.
 export const STAKE_ACCOUNT_PSEUDO_RENT_EXEMPT_RESERVE = 2_282_880
 
 // Mirrors `minimal_size_stake_account` of the validator-bonds program.
--- a/packages/validator-bonds-sdk/src/web3.js/stakeAccount.ts
+++ b/packages/validator-bonds-sdk/src/web3.js/stakeAccount.ts
@@ -267,6 +267,14 @@ export async function getRentExemptStake(
   )
 }
 
+// Pinned into every Meta by the stake program, which ignores it and takes the real floor from the Rent sysvar (SIMD-0490); bonds compares against the stored field, so never use the live rent here.
+export const STAKE_ACCOUNT_PSEUDO_RENT_EXEMPT_RESERVE = 2_282_880
+
+// Mirrors `minimal_size_stake_account` of the validator-bonds program.
+export function minimalSizeStakeAccount(minimumStakeLamports: BN): BN {
+  return minimumStakeLamports.addn(STAKE_ACCOUNT_PSEUDO_RENT_EXEMPT_RESERVE)
+}
+
 function pubkeyOrNull(
   value?: ConstructorParameters<typeof PublicKey>[0] | null,
 ): PublicKey | null {
--- a/settlement-pipelines/src/bin/close_settlement.rs
+++ b/settlement-pipelines/src/bin/close_settlement.rs
@@ -20,8 +20,8 @@ use settlement_pipelines::settlements::{
     load_expired_settlements, obtain_settlement_closing_refunds, SettlementRefundPubkeys,
 };
 use settlement_pipelines::stake_accounts::{
-    filter_settlement_funded, IGNORE_DANGLING_NOT_CLOSABLE_STAKE_ACCOUNTS_LIST,
-    STAKE_ACCOUNT_RENT_EXEMPTION,
+    fetch_stake_account_rent, filter_settlement_funded,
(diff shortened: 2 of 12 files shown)


settlement-pipelines/src/bin/close_settlement.rs
```diff
@@ -187,13 +187,17 @@ async fn close_settlements(
     reporting: &mut ReportHandler<CloseSettlementReport>,
 ) -> anyhow::Result<()> {
     let (bonds_withdrawer_authority, _) = find_bonds_withdrawer_authority(config_address);
+    let stake_account_rent = fetch_stake_account_rent(rpc_client.clone())
+        .await
+        .map_err(CliError::retry_able)?;
```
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — the fetch moved into the `split_rent_collector` branch of `obtain_settlement_closing_refunds`, so it only runs when a refund actually needs the value and the `stake_account_rent` parameter is gone. A failure there now reports per settlement and continues, as the sibling `collect_stake_accounts` error already did, instead of aborting `close_settlements` and `reset_stake_accounts`.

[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — the rent lookup now runs only when the batch actually contains a settlement with a `split_rent_collector`, so an empty or split-free batch makes no RPC call, and `CliError::retry_able` is kept so a transient failure still exits 100 for a Buildkite retry. Not caching it: that would drop N-1 constant-time `getMinimumBalanceForRentExemption` calls while leaving N uncached `collect_stake_accounts` program scans in the same branch.

packages/validator-bonds-sdk/src/web3.js/stakeAccount.ts
```diff
@@ -267,6 +267,14 @@ export async function getRentExemptStake(
   )
 }
 
+// Pinned into every Meta by the stake program, which ignores it and takes the real floor from the Rent sysvar (SIMD-0490); bonds compares against the stored field, so never use the live rent here.
```
kronael: I don't quite get this comment, nor the fact that we still need the constant.

settlement-pipelines/src/bin/close_settlement.rs
```diff
@@ -551,7 +564,7 @@ impl PrintReportable for CloseSettlementReport {
                 return vec!["No report available, not initialized yet.".to_string()];
             };
             let minimal_stake_account_lamports =
-                config.minimum_stake_lamports + STAKE_ACCOUNT_RENT_EXEMPTION;
```
kronael, line 554: Could we just - bump minimum_stake_lamports here, by say 1SOL, 10SOL, or 0.01SOL and remove the plus-handling?

kronael (dismissed):

packages/validator-bonds-sdk/src/web3.js/stakeAccount.ts
ochaloup: 🤖 claude: Fixed — the comment now reads plainly and names the two SIMDs separately. The constant stays: the program’s floor is the stored `Meta.rent_exempt_reserve`, still pinned at 2,282,880 (SIMD-0490) while live rent dropped to 2,077,224 (SIMD-0437), so a live-rent check would accept amounts the program then rejects.

settlement-pipelines/src/bin/close_settlement.rs
ochaloup: 🤖 claude: Declining — the program’s own floor is `stake_meta.rent_exempt_reserve + config.minimum_stake_lamports` (`utils/stake.rs`), so bumping the config field raises the real floor for every validator and the program still adds the reserve on top. The `+` would not go away, the reported number would just stop matching the chain.

settlement-pipelines/src/bin/close_settlement.rs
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Fixed — the lookup now runs only when the batch actually contains a settlement with a `split_rent_collector`, so an empty or split-free batch makes no RPC call, and `CliError::retry_able` is kept so a transient failure still exits 100 for the Buildkite retry. Per-settlement isolation is deliberately not taken: one blip aborting a batch is already capped by 5 retries on a once-per-epoch job.

ochaloup:
@kronael thanks, I updated comments to be hopefully better. :pray: re-review

settlement-pipelines/src/bin/close_settlement.rs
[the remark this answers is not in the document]
ochaloup: 🤖 claude: Also fixed the off-thread finding on this file — `set_closed_settlements` now receives only the settlements that actually got a close instruction, so one skipped after a failed refund lookup is reported as an error instead of counted as closed.

kronael (approved):

settlement-pipelines/src/bin/close_settlement.rs
kronael: Ok.  So is it so that the program is buggy?

ochaloup: Not a bug, it's intended - `rent_exempt_reserve` keeps the stake account alive and `config.minimum_stake_lamports` is the protocol minimum required on top of it, so bumping the config would raise the real floor for every validator while the program still adds the reserve.

